PR #53, opened 2014-12-07
In GNS3 run VirtualBox as another user

shmygov:
This code change allowed me to have VirtualBox VMs and a Cloud device with Internet in GNS3 network at the same time.
(Tested only on Ubuntu Linux)

I ran GNS3 as a root user (required to configure Cloud devices to use my real PC's network cards):
> sudo gns3server (in one terminal window)
> sudo gns3 (in another terminal window)
But with GNS3 running as root, I could not add my VirtualBox non-root VMs to GNS3, because they did not show up in GNS3 VMs list.

To solve this problem, I made the changes to GNS3 server and GNS3 GUI code adding the following new setting to GNS3:

Edit -> Preferences... -> VirtualBox -> General -> General settings -> "Run VirtualBox as another user (GNS3 running as Linux root)"

By default, this field is empty, and when it is empty GNS3 runs as usual, as if there were no code changes at all.
But if I input a non-empty user name and press Apply, all subsequent VirtualBox operations are performed in the name of this user, not the user who is currently running GNS3.

Now I can run GNS3 as Linux root, setup GNS3 to run VirtualBox as the user who owns the VMs, and add these VMs to GNS3 network.

Dmitry

### DIFF
--- a/gns3server/modules/virtualbox/__init__.py
+++ b/gns3server/modules/virtualbox/__init__.py
@@ -89,6 +89,8 @@ class VirtualBox(IModule):
         elif not os.access(self._vboxmanage_path, os.X_OK):
             log.warning("vboxmanage is not executable")
 
+        self._vbox_user = None
+
         # a new process start when calling IModule
         IModule.__init__(self, name, *args, **kwargs)
         self._vbox_instances = {}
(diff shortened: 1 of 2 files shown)


grossmj:
Manually merged.

PR #109, opened 2015-03-22
Added validator for uploading only image file

aka001:
Kindly review it.

### DIFF
--- a/gns3server/handlers/upload_handler.py
+++ b/gns3server/handlers/upload_handler.py
@@ -58,6 +58,14 @@ class UploadHandler:
             response.redirect("/upload")
             return
 
+        filename = data['file'].filename
+        file_format = filename.split('.')[-1]
+        valid_formats = ['jpeg', 'tif', 'png', 'gif', 'jpg', 'rif', 'bmp', 'bpg']
+
+        if file_format not in valid_formats:
+            response.redirect('/upload')
+            return
+
         if data["type"] not in ["IOU", "IOURC", "QEMU", "IOS"]:
             raise aiohttp.web.HTTPForbidden("You are not authorized to upload this kind of image {}".format(data["type"]))
 


noplay:
Hi,

You PR has an issue.

The "image" are IOU, Qemu or VirtualBox. It's not an image like a JPEG but the snapshot of the hard drive of an OS.

aka001:
@noplay Okay. I get it. Can we verify this by checking on extension of the file? We can also check the size of the uploaded file for this. Atleast some part of data will be filtered.

gns3server/handlers/upload_handler.py
```diff
@@ -58,6 +58,14 @@ def upload(request, response):
             response.redirect("/upload")
             return
 
+        filename = data['file'].filename
+        file_format = filename.split('.')[-1]
+        valid_formats = ['jpeg', 'tif', 'png', 'gif', 'jpg', 'rif', 'bmp', 'bpg']
+
+        if file_format not in valid_formats:
+            response.redirect('/upload')
+            return
```
rashoodkhan: I think it would be better to return HttpResponse with an error so the user knows he made a mistake while uploading the file. Error message explaining what happened would be informative to the user.

grossmj:
I agree that we should check what can be accepted by the Upload handler. I already have code to detect IOS and IOU images by reading the ELF header.

IOS: https://github.com/GNS3/gns3-server/blob/153914bf9712d267d13769c4a6bee2dd86a1cdce/gns3server/modules/dynamips/nodes/router.py#L257

IOU: https://github.com/GNS3/gns3-server/blob/153914bf9712d267d13769c4a6bee2dd86a1cdce/gns3server/modules/iou/iou_vm.py#L161

However, the problem are QEMU images, I don't know how we could validate that.

gns3server/handlers/upload_handler.py
noplay: Yes you are right

noplay:
I fear that qemu as a lot possibility:
http://en.wikibooks.org/wiki/QEMU/Images

aka001:
I discussed with @grossmj about this. We need to put a validator for qcow2 type of image files. 
From this link https://people.gnome.org/~markmc/qcow-image-format.html I found out that a qemu file can be identified by the property that it's first 4 bytes contain the characters 'Q', 'F', 'T' followed by 0xfb. 
Kindly mention if I am going wrong here.
We can also integrate @grossmj's code in it to identify IOS and IOU files which will make it a complete checker.

ehlers:
Limiting it to qcow2 makes it rather inflexible.

For qemu I just can think of the following file types:
raw, qcow2, vmdk, vdi, iso, linux image, linux initrd

I'm sure I have forgotten something. Do you really want to check for all these file type ?
I would suggest "junk in - junk out", meaning the user has to know what he is doing.

aka001:
@ehlers , IMHO, we must add these checkers.
There might be a use case in which user adds invalid file (doesn't match any of the file format) unaware of the fact that it's an invalid file and might think that the functionality of gns3-server is broken .

noplay:
Perhaps we should not check the file type depending of what the user choose. And for qemu we can allow anything.

PR #149, opened 2015-04-10
Use hex encoding on VirtualBox pipe names

boenrobot:
This allows arbitrary characters in VirtualBox VM names, while also ensuring all different weird names can coexist.

Example of two VMs that previously couldn't both have a serial console - "VM 1" and "VM_1", because the former's pipe name is converted to the latter, while the latter would remain unchanged.

**P.S. NOT TESTED AT ALL.**

### DIFF
--- a/gns3server/modules/virtualbox/virtualbox_vm.py
+++ b/gns3server/modules/virtualbox/virtualbox_vm.py
@@ -583,8 +583,7 @@ class VirtualBoxVM(BaseVM):
         :returns: pipe path (string)
         """
 
-        p = re.compile('\s+', re.UNICODE)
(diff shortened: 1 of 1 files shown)


grossmj:
@boenrobot thanks for proposing this change. Converting to hex didn't worked so I've done it differently but the result is the same.

Please see https://github.com/GNS3/gns3-server/commit/161adb781b84b765d96106c5e43de4b45f7e3580

grossmj:
IDs are just UUIDs:  http://en.wikipedia.org/wiki/Universally_unique_identifier

Only hex numbers separated with -

So no possibility of forbidden characters :)

PR #154, opened 2015-04-21
Support wireshark remote capture

noplay:


gns3server/handlers/api/file_handler.py
```diff
@@ -0,0 +1,54 @@
...
+                response.content_type = "application/octet-stream"
+                response.set_status(200)
+                response.enable_chunked_encoding()
+                # Very important: do not send a content lenght otherwise QT close the connection but curl can consume the Feed
+                response.content_length = None
+
+                response.start(request)
+
+                while True:
+                    data = yield from loop.run_in_executor(None, f.read, 16)
```
grossmj: I have only little concern about using executors, what happen when we hit max_workers limit which I think is 4 by default? According to docs a task waits for an executor to become available, this probably means that we can have a maximum of 4 concurrent captures.

noplay: On my mac i successfully run 111 capture in parallel. I was block by the too many open files.  I will push a second commit with an explicit ThreadPoolExecutor of 100 thread. Do you think we need more?

grossmj: 100 is fine, I doubt that a user would need more than a few captures at the same time.

PR #158, opened 2015-04-24
Do not erase the IOU config if you already have one

noplay:
I have try to do the same with dynamips but this part is not tested and it's deeply in the code and I fear to introduce a side effect because I can't do it in the handler.

From what I understand of the code we overwrite it at each create. It's the client which protect us against this.


noplay:
It's mergeable

PR #167, opened 2015-04-30
Upload images API and upgrade to aiohttp 0.15.3 

noplay:


dlintott:
Is it necessary to have such a strict dependancy on the version of aiohttp?

Currently debian has 0.15.1, so unless a specific feature is needed in 0.15.3, these requirements should be the lowest possible that work

noplay:
Ok for 0.15.1

noplay:
@dlintott we will try to match as close  as possible Debian requirements.

PR #168, opened 2015-05-01
A notification stream with process monitoring

noplay:


noplay:
Be  carefull to destroy your local branch before checking out this. After the merge issue you need to use a clean branch.

PR #184, opened 2015-05-08
Readme + setup.py/requrements.txt sync

boenrobot:
Similar to GNS3/gns3-gui#321 in purpose - the two files are simply not in sync.

Is python-dateutil needed btw?

I erred on the side of including it, but I honestly have no idea... a search of "dateutil" doesn't show any occurrences... Should it be removed from both places instead of being added to both?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,16 +20,17 @@ Branches
 master
 ******
 master is the next stable release, you can test it in your day to day activities.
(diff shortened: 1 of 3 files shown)


grossmj:
Hi Vasil,

You are right, we do not use dateutil :+1:  Please remove it.

Thanks,

grossmj:
Merged, thanks :)

PR #187, opened 2015-05-11
Dedicated Qemu-img method and API call

boenrobot:
This is a prerequisite for a Qemu-img wizard, which searches for the Qemu-img binary on the local server using this method, to then invoke it.

The alternative is to duplicate the path finding code into gns3-gui, but that's IMHO unnecessary and just not a good idea.

### DIFF
--- a/gns3server/modules/qemu/__init__.py
+++ b/gns3server/modules/qemu/__init__.py
@@ -39,14 +39,13 @@ class Qemu(BaseManager):
     _VM_CLASS = QemuVM
 
     @staticmethod
-    def binary_list():
+    def paths_list():
         """
-        Gets QEMU binaries list available on the host.
+        Gets a folder list of possibly available QEMU binaries on the host.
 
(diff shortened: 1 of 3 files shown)


gns3server/modules/qemu/__init__.py
```diff
@@ -75,7 +74,18 @@ def binary_list():
             paths.extend(["/usr/local/bin", "/opt/local/bin"])
```
noplay: I think we should add /usr/bin

noplay: Forgot my comment it's only OSX

noplay:
It could be nice if we add tests.  Let me know if you need help.

I have fixed the tests on the unstable branch. If you merge or rebase the PR should be green.

noplay:
Yeah servers side is easy to tests :)

grossmj:
Thanks for your patience :)

PR #239, opened 2015-06-18
Make VMX reading case insensitive

### LINKED ISSUE #237: /vmware/vms doesn't work on OSX
It's prevent usage of GNS3 VM

noplay:
Fix #237

I lower all the key in VMX file. But I fear to broke something


grossmj:
VMware writes camel case keys in VMX files but this is not the case when a VM is built using packer. I agree it is safer the just have all keys lower case.

grossmj:
Well main issue now is all VMware files we use have lower case keys (because we write back to them)... which is a bit ugly.

grossmj:
Well, I can live with this ;)

PR #271, opened 2015-07-20
Update gns3.conf.upstart

lykinsbd:
Fixing start-stop-daemon syntax so that it accepts the arguments correctly on Ubuntu 14.04.

As it stands, starting the process fails with the following message because it tries to directly interpret "--log" as an option of start-stop-daemon, not an argument applied to gns3server:

start-stop-daemon: unrecognized option '--log'
Try 'start-stop-daemon --help' for more information.

### DIFF
--- a/init/gns3.conf.upstart
+++ b/init/gns3.conf.upstart
@@ -6,7 +6,7 @@ stop on shutdown
 
 script
     echo $$ > /var/run/gns3.pid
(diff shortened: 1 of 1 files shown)


grossmj:
Thanks a lot :+1:

lykinsbd:
Glad to help!

Issue #358, opened 2015-11-13
TCP port 2000 is outside the range 2001-7000


ehlers:
GNS3 version 1.4.0rc1 on Darwin (64-bit) with Python 3.5.0 Qt 5.5.0.

I have an old project, that uses console port 2000. Starting that gives
```
Server error from http://192.168.1.90:8000: IOU1: TCP port 2000 is outside the range 2001-7000
```

Manually changing the console port to 2001 in the .gns3 file resolves it. A grep shows, that this is the only project with console port 2000, all other start with 2001. I assume the fix https://github.com/GNS3/gns3-server/commit/bd47c21514177fdfef44888f09b478f6f405d5c5 causes that.

Update: This user is also affected: https://community.gns3.com/thread/22938#51156

ehlers:
This project uses a remote server with a minimal gns3_server.conf, besides others console_start_port_range was not defined in this config file. Before change https://github.com/GNS3/gns3-server/commit/bd47c21514177fdfef44888f09b478f6f405d5c5 the default console_start_port_range was 2000, therefore it was used as a console port for projects using this server. With the new minimum start port of 2001 a project using the old port 2000 is rejected.

noplay:
@grossmj do you have a specific reason for increasing to 2001?

grossmj:
The original intend was that R1 = 2001, R2 = 2002 etc.

I think we should not check the range when loading a project.

noplay:
But if we turn off the check and you have only open some port in firewall
you will not get an error.

Otherwise we can start router at R0 :p

Le sam. 14 nov. 2015 00:33, Jeremy Grossmann <notifications@github.com> a
écrit :

qianxideyu:
Server error from http://192.168.56.101:8000: P1: TCP port 2000 is outside the range 2001-7000，in gns3 default config ,the default port range 2001 to 7000!

qianxideyu:
On windows，GNS3 1.4RC`s vmware,can stop,onlu delete it ,If II click the  stop button ,it will time out 60s!

noplay:
I have sent a PR #370  the idea is to never trust the client for the console port. This should solve a lot of issues if you change the settings of the port range or move a topology to another computer.

grossmj:
Resolved by https://github.com/GNS3/gns3-server/pull/374

PR #370, opened 2015-12-02
Always affect a free console port

noplay:
Instead a leaving the choice to the client this always
affect a random free console port. This should solve
issues when two users are on the same server or
you opened GNS3 twice.

I'm not happy to make a big change like this now... But it create so much trouble... 

I still accept a console send by client but ignore it. Because I don't want to touch the client during the RC process on this.

### DIFF
--- a/gns3server/handlers/api/dynamips_vm_handler.py
+++ b/gns3server/handlers/api/dynamips_vm_handler.py
@@ -70,7 +70,6 @@ class DynamipsVMHandler:
                                                    request.json.get("vm_id"),
                                                    request.json.get("dynamips_id"),
                                                    platform,
-                                                   console=request.json.get("console"),
                                                    aux=request.json.get("aux"),
                                                    chassis=request.json.pop("chassis", default_chassis))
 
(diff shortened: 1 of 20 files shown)


noplay:
This solve #358

grossmj:
I see only one problem, it is that some users want to control what port number is used for a particular device. Besides this I thought that if a port is not available we automatically pick another one.

noplay:
Actually do you have a way to force the port in the GUI?

grossmj:
Yep you can do it in the GUI

grossmj:
There is an auxiliary console port for Dynamips IOS router named "aux". The PR should cover that as well.

noplay:
We can change the PR for accepting a port. And switching to another in case of problem. 

But I don't like the the idea of the user getting the correct port 99% but sometimes magically we change the port of the user.

And changing the port in the GUI will work for local usage, but with multi users and portability the features will be broken.

The ideal solution could be to have a nothing a dynamic console port in the GUI and we force a port and raise error in case of issues only if user choose to manually force a port. Problem it's require changing a lot in the GUI and we have no idea in old topologies if user has change the port or not.


An intermediate solution is to let the magic happen but send a warning to the client. In rare case the user could be angry if he choose to force port, and the port are not free he will lost his settings if he save the topology.

PR #374, opened 2015-12-07
Replace by another TCP port if port is already used

noplay:
Another version of #370

This time we replace a free TCP port if port is used and raise
a warning to the user.

capture d ecran 2015-12-07 a 12 26 31

### DIFF
--- a/gns3server/modules/base_vm.py
+++ b/gns3server/modules/base_vm.py
@@ -60,11 +60,14 @@ class BaseVM:
         self._vm_status = "stopped"
 
         if self._console is not None:
-            self._console = self._manager.port_manager.reserve_tcp_port(self._console, self._project)
+            if console_type == "vnc":
+                self._console = self._manager.port_manager.reserve_tcp_port(self._console, self._project, port_range_start=5900, port_range_end=6000)
+            else:
+                self._console = self._manager.port_manager.reserve_tcp_port(self._console, self._project)
         else:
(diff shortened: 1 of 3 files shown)


ehlers:
With an (local) IOS router, I get this warning in the GUI that the port has changed. But on an (remote) IOU router, I don't get this warning in the console. The warning is on the remote server log und when I hover over the router I see the changed console port. But I'm not warned in the console.

noplay:
Strange because the notification is send via the HTTP notification stream....

ehlers:
As a test I created a topology with 2 dynamips router and then manually edited in the gns3 file the console port to 2001 for both routers.

When I activate debug logging on my local server (for IOS) I see 
```
2015-12-10 11:14:59 WARNING port_manager.py:230 TCP port 2001 already in use on host 192.168.1.10. Port has been replaced by 2002
```
and later
```
2015-12-10 11:14:59 DEBUG project_handler.py:216 Send notification: {"action": "log.warning", "event": {"message": "TCP port 2001 already in use on host 192.168.1.10. Port has been replaced by 2002"}}
```

I litte bit strange is, that I get this two times, but I don't want to be too picky.

When I do the same with IOU I get
```
2015-12-10 10:15:02 WARNING port_manager.py:230 TCP port 2001 already in use on host 0.0.0.0. Port has been replaced by 2002
```
also two times, but no sign of sending a notification.

PR #386, opened 2016-01-01
Check KVM support

boenrobot:
Adds an API point to check if the server supports KVM. Returned as part of an object of "capabilities", that currently includes just KVM.

It's defined in this way to minimize "feature creep" - if similar functionality is later required for other system related Qemu checks, they should just be added as part of this response's output.

### DIFF
--- a/gns3server/modules/qemu/__init__.py
+++ b/gns3server/modules/qemu/__init__.py
@@ -21,6 +21,7 @@ Qemu server module.
 
 import asyncio
 import os
+import platform
 import sys
 import re
 import subprocess
@@ -37,6 +38,27 @@ log = logging.getLogger(__name__)
 class Qemu(BaseManager):
 
     _VM_CLASS = QemuVM
+
+    @staticmethod
+    def get_kvm_archs():
+        """
+        Gets a list of architectures for which KVM is available on this server.
+
+        :returns: List of architectures for which KVM is available on this server.
+        """
+        kvm = []
+        x86_64_aliases = ["x86_64", "x86-64", "x64", "AMD64", "amd64", "Intel 64", "EM64T"]
+        i386_aliases = ["i386", "x86", "x32"]
+        if sys.platform.startswith("linux") and subprocess.call("kvm-ok") == 0:
+            arch = platform.machine()
+            if arch in x86_64_aliases:
+                kvm.append("x86_64")
+                kvm.append("i386")
+            elif arch in i386_aliases:
+                kvm.append("i386")
+            else:
+                kvm.append(platform.machine())
+        return kvm
 
     @staticmethod
     def paths_list():
@@ -82,7 +104,7 @@ class Qemu(BaseManager):
         return paths
 
     @staticmethod
-    def binary_list():
+    def binary_list(archs=None):
         """
         Gets QEMU binaries list available on the host.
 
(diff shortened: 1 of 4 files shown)


noplay:
Thanks for the proposal.

Before merging it we should add a : 
* a test of the call (this will allow to generate the doc)
* a schema for the answer (schema is here: https://github.com/GNS3/gns3-server/blob/master/gns3server/schemas/qemu.py and for using it https://github.com/GNS3/gns3-server/blob/master/gns3server/handlers/api/qemu_handler.py#L47)
* improving the KVM check

About the check of KVM it's more complicated. We need to call the kvm-ok command to check if KVM is really available: https://github.com/GNS3/gns3-vm/blob/master/scripts/welcome.py#L200

Because you have multiple case for non KVM support:
* not a linux
* kvm module not installed
* cpu doesn't support VT instructions

noplay:
I'm not sure if kvm-ok is always install on computer with KVM

noplay:
Perhaps we could return the KVM support with the list of KVM binaries and when installing an appliance checking also that the binary is available.

noplay:
KVM i'm sure at 95% that KVM will work for i386 on a 64 bit host.

After checking it seem to be the case:
http://www.linux-kvm.org/page/Guest_Support_Status


I don't know if you have performance gain for qemu-arm on a x86 .

And KVM could even work on some specific arm processor:
http://wiki.qemu.org/KVM

noplay:
No worry :P Contributions are precious and we really thank you for your work.

I will test it soon.

gns3server/modules/qemu/__init__.py
```diff
@@ -39,6 +40,27 @@ class Qemu(BaseManager):
...
 
     @staticmethod
+    def get_kvm_archs():
+        """
+        Gets a list of architectures for which KVM is available on this server.
+
+        :returns: List of architectures for which KVM is available on this server.
+        """
+        kvm = []
+        x86_64_aliases = ["x86_64", "x86-64", "x64", "AMD64", "amd64", "Intel 64", "EM64T"]
```
noplay: From where the list come from?

noplay: I think GNS3 allow only x86_64

boenrobot: GNS3 and Qemu allow only x86_64, but that very same architecture is identified in different ways by different OS-es, as is i386, so to match all names used by OS-es, the OS value is matched against this list, before being forced into the name that Qemu uses.

The list came from [the Wikipedia article on x86_64](https://en.wikipedia.org/wiki/X86-64), and primarily the "Industry naming conventions" sections. That list doesn't include the two Intel names - "Intel 64" and "EM64T" - but they're official, so they're added just in case this article is missing some fringe Linux flavor that actually uses them. I mean, you can see even different Linux flavors use other different names. The list also includes the Windows and OSX names, even though they don't currently have KVM support, and that is just in case they do OR (more likely) if a Linux flavor tries to make itself compatible with them by adopting their naming convention.

noplay: Due to the fact we control the input files via the registry we can be strict and keep only one name by architecture I think. I don't see problem with what you have propose but I fear to have trouble in the future.

gns3server/modules/qemu/__init__.py
```diff
@@ -39,6 +40,27 @@ class Qemu(BaseManager):
...
+        i386_aliases = ["i386", "x86", "x32"]
+        if sys.platform.startswith("linux") and subprocess.call("kvm-ok") == 0:
```
noplay: We will need to use an async subprocess call otherwise the server will freeze until kvm answer

noplay:
I take care of the PR

noplay:
It's cleaned and merged

gns3server/modules/qemu/__init__.py
boenrobot: You don't control the architecture name returned from platform.machine() though.

gns3server/modules/qemu/__init__.py
boenrobot: You don't control the architecture name returned from platform.machine() though.

noplay:
After reading the python source code you are right it's not working on freebsd and openbsd. Because it's use the output of uname.
https://en.wikipedia.org/wiki/Uname

Ubuntu use linux convention which is x86_64

noplay:
What is the status of KVM under Freebsd? From what I read the project is dead and has never really been utilized in production.

Since bhyve enter in the game I think the community has now a good solution.

jean-christophe-manciot:
One way to check for Intel VT or AMD-V CPU virtualization capabilities without "kvm-ok" is:
egrep '(vmx|svm)' --color /proc/cpuinfo, 1 line / vCPU

jean-christophe-manciot:
It's even easier to count the number of vCPU:
**sudo egrep -c '(vmx|svm)' /proc/cpuinfo**

jean-christophe-manciot:
It works without sudo:
actionmystique@msi-ge60-ubuntu:~$  egrep -c '(vmx|svm)' /proc/cpuinfo
8

noplay:
@jean-christophe-manciot  the CPU can support VT instruction but the KVM modules can be missing.

But it's interesting we can can add this information in order to explain to the user that his computer CPU is not supported by KVM even if he add the KVM module

PR #418, opened 2016-02-03
Do not list qemu binary with -spice in the name

noplay:
It's create confusion and you can enable spice with the -spice
options.

Fix https://github.com/GNS3/gns3-gui/issues/981


grossmj:
Not 100% comfortable about filtering this but anyway, let's see.

noplay:
Agree

PR #448, opened 2016-03-03
updated docker command line processing

ehlers:
Made the following changes:
- use image_infos.get("Config",...) instead of "ContainerConfig"
- init script injected before normal Entrypoint, so it's guaranteed to be the first command running
- init script has chmod +x, started without /bin/sh, because init script needs pid 1
- init script starts next script/program with exec, because (if it's a "real" init) it needs pid 1
- command line in GNS3 replaces command in docker, it's not appended. That's consistent with the "docker run" behavior
- A shell /bin/sh is only added, when Entrypoint and Cmd is empty

### DIFF
--- a/gns3server/modules/docker/docker_vm.py
+++ b/gns3server/modules/docker/docker_vm.py
@@ -195,16 +195,22 @@ class DockerVM(BaseVM):
             },
             "Volumes": {},
             "Env": [],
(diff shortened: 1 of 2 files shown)


noplay:
I read it and it seem good. I will take time to test it tomorrow when I will not be in the middle of something else.

Thanks a lot !

noplay:
It's perfect merged !

PR #460, opened 2016-03-23
Docker init supports /etc/network/interfaces

ehlers:
As discussed in https://github.com/GNS3/gns3-gui/issues/1089 and https://github.com/GNS3/gns3-gui/issues/1097 it's an option to support /etc/network/interfaces in the GNS3 specific docker init.

In this PR that's done by injected it in the docker container through the INTERFACES environment variable.

Here an example for a static configuration:
`INTERFACES=auto eth0\niface eth0 inet static\n\taddress 10.1.1.100\n\tnetmask 255.255.255.0\n\tgateway 10.1.1.1\n\tup echo nameserver 127.0.0.1 > /etc/resolv.conf`

And here a DHCP configuration:
`INTERFACES=auto eth0\niface eth0 inet dhcp`

This is obviously not a reasonable user interface. There must be some user friendly way in the GNS3 GUI to enter the required data, create a file in the etc/network/interfaces syntax and mount that into the docker container.

It uses an embedded busybox binary to be independent of the binaries available in the docker container.

Open questions:
- Do we want to go this way?
- Somebody has to define and implement a reasonable user interface
- Do we want to optimize the busybox binary? Currently I'm using the alpine busybox-static package, it's about 900kb. By creating a binary with only the needed features we can save some space.
- Currently all interfaces are activated, even when they are not defined in /etc/network/interfaces. This is not conform to Debian's interface handling, but eases the use of those interfaces in bridged environments (e.g. Open vSwitch). Do want to keep that?
- The delay of 2 seconds for activating the interfaces is copied from https://github.com/GNS3/docker-ovs . Is it necessary ?


ehlers:
Forgot to mention, that resources/etc/udhcpc/default.script is copied from the alpine busybox-initscripts package. Only some file paths have been changed.

noplay:
It seem we are in the good way.

We can ship a very simple interface. 

By default DHCP is enabled on all interfaces. Plus for all interfaces a commented version for the static.
In the docker preferences you can edit the generated interfaces.

But I suggest we create by default a volumes /etc/network and put this file in it. And mount this volume by default. And I provide a way a simple way to edit the file from the interface (an edit interfaces text area where you edit and when you save it's pushed back on server). It's probably more futur proof.

About busybox. My suggestion if you agree @grossmj we create a git with busybox and we build our own version. I remember that is easy just make menuconfig and make. And hosting the source somewhere will allow us to avoid licence issue (https://www.busybox.net/license.html) and track what is enable in our binary.

But for distribution we keep the busybox bin in ressources to make install simple.

Personnaly I prefer when everything is enabled. It's avoid to search why it's not working.

For the delay I think it's because we need time for ubridge to inject all the network interfaces in the container. A better way could be that GNS3 pass the number of adapter require via an ENV var and in the init we loop until they are available (we can just wait the last one).

For your informations we think we will ship a 1.5 with docker and move bigger changes (on the networking) in a version named 2.0 for the moment. Allowing us to ship quickly the docker support.

ehlers:
Sounds good.

ehlers:
Regards building busybox:

Created https://github.com/ehlers/busybox/tree/gns3_build . It's a fork of https://github.com/mirror/busybox at tag 1_24_2. I added a subdirectory gns3_build, that (what a surprise) builds a static busybox for use in the docker init. I copied almost everything from the alpine busybox build environment, because they have a lot of experience with it.

The only GNS3 specific modification is a changed default udhcpc script location. That way the udhcp shell wrappers in my commit https://github.com/GNS3/gns3-server/pull/460/commits/d36a9f34031eb2093afb69c1ca7efadd643ea4bd are no longer necessary.

noplay:
Thanks it's forked and I add you to the project:
https://github.com/GNS3/busybox

ehlers:
Thanks, just did some bug fixes. If you already created a binary, please do it again.

ehlers:
About using DHCP as default in network/interfaces:

In the current busybox implementation of network/interfaces on each DHCP interface it tries to get a lease 5 times. If it doesn't get a lease, busybox will give up on that interface, no dhcp client is started in the background. By adding the -b flag it should be possible that udhcpc backgrounds then.

Nevertheless we get a long delay on startup if we have several links, where no DHCP server is active. As I assume that this is the default in a GNS3 topology, I suggest  to have the dhcp protocol also commented in the interfaces file, so the user has to uncomment either DHCP or static to have an IP configuration.

noplay:
You are right

noplay:
Interface added for the client:
https://www.youtube.com/watch?v=dRgPt4wtOmE

ehlers:
Super, docker VMs are really usable now :smiley: Ready for a release candidate ?

grossmj:
I think we will release an alpha soon.

PR #467, opened 2016-03-30
Minor cleanup on docker-init

ehlers:
Some minor cleanup
- avoid warnings by adding /etc/network/pre-up.d and post-down.d
- shell wrappers for udhcpc and udhcpc6 no longer necessary, script path has been changed in busybox

### DIFF
--- a/gns3server/modules/docker/docker_vm.py
+++ b/gns3server/modules/docker/docker_vm.py
@@ -181,6 +181,8 @@ class DockerVM(BaseVM):
         os.makedirs(path, exist_ok=True)
         os.makedirs(os.path.join(path, "if-up.d"), exist_ok=True)
         os.makedirs(os.path.join(path, "if-down.d"), exist_ok=True)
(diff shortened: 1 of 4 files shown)


ehlers:
The tests on 1.5 are failing since this commit:
1.5 New console port range 5000 to 10000.

ehlers:
Just one more idea:

Currently the busybox ifupdown gives up getting a lease after 5 tries. It's the option `-n` of udhcpc, which is responsible for that. When using `-b` instead, it should continue to try to get a lease in the background (according to the documentation, I haven't tried).

To change that in the config file of the busybox build
`CONFIG_IFUPDOWN_UDHCPC_CMD_OPTIONS="-R -n"`
should be replaced by
`CONFIG_IFUPDOWN_UDHCPC_CMD_OPTIONS="-R -b"`

noplay:
I pushed a fix on 1.5 for the test failed

noplay:
Thanks !

PR #470, opened 2016-03-30
Import export API

noplay:
This provide an API for exporting a project directory as ZIP and importing it via the API.

For the moment I called this archive a gns3z (Zipped GNS3) but we probably will use a better name.

The export will ignore:
* snapshots
* logs files

Latter we will need to improve qemu to support file relocation.

The export code in the GUI is ready it's relatively small:
https://github.com/GNS3/gns3-gui/blob/46b0012e5c457d50ec890b89fed2f055d2f1cd87/gns3/utils/export_project_worker.py

The only missing part is the import before pushing a PR.

Ref: https://github.com/GNS3/gns3-gui/issues/476

### DIFF
--- a/gns3server/handlers/api/project_handler.py
+++ b/gns3server/handlers/api/project_handler.py
@@ -20,6 +20,8 @@ import asyncio
 import json
 import os
 import psutil
+import tempfile
+import zipfile
 
 from ...web.route import Route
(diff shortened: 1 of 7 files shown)


noplay:
The snapshots are not included to save disk space and simplify operations.

PR #479, opened 2016-04-05
Randomize port assignation

### LINKED ISSUE #1154: Nexus 9k Qemu appliance PCI: no slot/function available for e1000, all in use
https://gns3.com/qa/nexus-9k-qemu-appliance-pci-no-s

noplay:
Actually port are sequential this mean it's easy to got
conflict between projects or with different usage (example:
docker aux console is not saved to disk).

Fix #1154

### DIFF
--- a/gns3server/modules/port_manager.py
+++ b/gns3server/modules/port_manager.py
@@ -17,6 +17,7 @@
 
 import socket
 import ipaddress
+import random
 from aiohttp.web import HTTPConflict
 from gns3server.config import Config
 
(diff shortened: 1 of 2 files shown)


grossmj:
Some users like when the ports are allocated sequentially, e.g. R1 = 2001, R2 = 2002 since they can use terminals outside GNS3. However, I agree I think this won't be possible to maintain that  if we want to support multiple users.

noplay:
Or multiple projects opened. Also if we use the controller has a proxy we will have conflict between ports of local server and port of the VM I think VIRL do that for their large topo on multiple packet machine.

grossmj:
For 1.5 we will save Docker AUX ports in .gns3 files, however for 2.0 we will re-factor and not save console ports in topologies (just like we do already for UDP ports).

Issue #515, opened 2016-05-06
Parse version is not perfect


noplay:
See #513

grossmj:
Fixed

PR #516, opened 2016-05-06
Improve parse_version

noplay:
What  do you think @grossmj ?  I think It should support most semantic version comparaison (problem you should not use the output to display it to a user)

I could backport it to GUI if it's ok

Ref #515


grossmj:
Looks like it works. I'll backport to the GUI.

PR #518, opened 2016-05-06
Wait for all interfaces available in container

### LINKED ISSUE #511: ifup: can't open '/var/run/ifstate': No such file or directory
If a controller is slow to start (due to High CPU usage in the GNS3 VM) we got this error

noplay:
Proposal improve for waiting that all the eth interface are available in the container.

Fix #511


noplay:
Could interrest @elhers

PR #630, opened 2016-07-29
Move utils.vmnet to gns3 namespace

athmane:
Hi,
vmnet should be moved to gns3 namespace to avoid conflict with other packages.
This will also help with disto packaging (in my case Fedora).


noplay:
Hi,

Thanks a lot  we will check that the week end

PR #648, opened 2016-08-25
Updated systemd unit file and added sample configuration file

ianc1215:
Rewrote the systemd unit file to fix an issue where the system was not able to create a PID file inside /var/run.

I fixed this by having systemd create a new directory called /var/run/gns3. Then I had systemd change ownership of the directory to gns3:gns3 so the gns3server executable could read and write the PID file. I have tested these changes against Ubuntu 16.04.1 LTS.


noplay:
Thanks  a lot!

PR #653, opened 2016-08-27
Update 'Updated systemd unit file and added sample configuration file'

ehlers:
Changes CR/LF line ending to LF only.

conf/gns3_server.conf
- set console_start_port_range, console_end_port_range and auth to default values
- comment out paths, as they have no fixed default

Undecided, therefore not changed:
`;` is used as comment char, maybe `#` might be more consistent with other example files

init/gns3.service.systemd
- added option `-p`to mkdir, otherwise mkdir will fail, when a directory (/var/run/gns3) already exists. That will lead to a failure of `service gns3 start` (at least in debian).
- additionally mkdir/chown /var/log/gns3, this directory is needed as well.
- correct PIDFILE, otherwise `service gns3 start` waits forever for the wrong pid file (at least in debian)


noplay:
Thanks !

PR #726, opened 2016-10-18
When checking for a free port check if the host and 0.0.0.0 are available

### LINKED ISSUE #721: Check how we verify that a UDP port is busy
https://gns3.com/qa/unable-to-create-udp-nio-mac-mav (at the bottom of the discussion)

noplay:
Because some emulators will listen on 0.0.0.0 and not on the host.

Fix #721


grossmj:
Looks ok but be on the lookout for possible side effects.

PR #856, opened 2017-01-06
Change directory layout for dynamips.

noplay:
This create a directory by node to allow feature like save as
to work when running on multiple compute. Benefits now dynamips work
like other nodes. Old topologies are migrated at startup.

Ref https://github.com/GNS3/gns3-gui/issues/1761


ehlers:
Conversion from old directory layout to the new one missing for devices running on remote servers.

noplay:
Ok should work on remote now

ehlers:
Tested with remote dynamips devices, works fine.

noplay:
Thanks ! As usual your help is priceless.

PR #890, opened 2017-02-03
Manage base configuration on server

### LINKED ISSUE #786: Manage base configuration on server
Actually the config file could be located anywhere on the client system. This lead to multiple issue. 
* Two client can send different base config file.
* Complex API call where sometimes you need to pass the base config file content to the call (at creation) but not later
* If you create a third partie client he will not use the default config file

Todo:
* [x] Move creation of base file if not exists from client to server
* [x] Remove sending base file from client to server
* [x] When creating a node that require a base file send it from controller to client
* [x] Support abs and non abs paths
* [x] Rewrote config import / export to not use the node object but dedicated API
* [x] Disable file browser in GUI for remote controller

noplay:
Server side of https://github.com/GNS3/gns3-gui/pull/1835

This is a big cleanup. The default configs are now manage on server.

This:
* clean a lot of code for supporting that on client without erasing the remote config by mistake
* support central location for default base config when use in multiple client context
* support export config for docker when we export all configs

Fix #786

### DIFF
--- a/docs/file_format.rst
+++ b/docs/file_format.rst
@@ -23,6 +23,7 @@ A minimal version:
 
 The revision is the version of file format:
 
+* 8: GNS3 2.1
 * 7: GNS3 2.0
 * 6: GNS3 2.0 < beta 3
 * 5: GNS3 2.0 < alpha 4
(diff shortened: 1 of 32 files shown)


grossmj:
I thought I had detected a problem with this PR regarding private-config but it also happens in 1.5.x and 2.0: https://github.com/GNS3/gns3-server/issues/906

PR #979, opened 2017-04-15
typing is already included in Python >= 3.5

athmane:
typing is now part of Python starting from 3.5, this PR exclude this requirement on such version.


noplay:
Thanks!

PR #987, opened 2017-04-27
NAT node can use the VMware NAT

noplay:
On Windows and Mac OS this allow the NAT node to use the
VMware nat (it's always vmnet8). On Linux we still use the
libvirt NAT


grossmj:
This will need to be tested on Windows.

noplay:
For the moment on windows it will not work:
https://github.com/GNS3/gns3-server/issues/989

PR #1031, opened 2017-05-15
When connecting to a compute node ask for qemu images list

### LINKED ISSUE #996: Timeout when loading large image library
Hi again,

This is for 2.0 using a remote server.  The first time an image (qemu etc)  is configured on 2.0 and the image directory has a lot of images (for example ours has about 40gb of images) the client will timeout waiting for http://10.x.x.x:3080/v2/computes and the server will stay at 100% CPU for a while.  In this case, the client does not give the server enough time to process the images and aborts the process.  The only way to complete the process was to run  http://10.x.x.x:3080/v2/computes from a browser and let the server finish up.

Does this make sense?  Let me know if I need to clarify.

noplay:
The server start to compute qemu image list in background
because first time discovery could be slow due to the compute
time of md5sum.

Fix #996

### DIFF
--- a/gns3server/web/web_server.py
+++ b/gns3server/web/web_server.py
@@ -33,6 +33,7 @@ from .route import Route
 from ..config import Config
 from ..compute import MODULES
 from ..compute.port_manager import PortManager
+from ..compute.qemu import Qemu
 from ..controller import Controller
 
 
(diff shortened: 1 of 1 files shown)


noplay:
I'm not a big fan, but I don't see alternatives. 

In theory you should upload images to GNS3 using GNS3 to avoid this kind of troubles. But some users prefer to manage image directly on the server (and worse on a network share).

Issue #1000, opened 2017-05-03
Upgrade to aiohttp 2.0


noplay:

noplay:
http://aiohttp.readthedocs.io/en/stable/migration.html

noplay:
The migration progress but it's harder than expected.

Now the server seem to work but two test fail. It seem a problem with reading multiple chunk of a response in tests, when the same thing work from a browser.

noplay:
The good news is the problem is the same on all Python version or on any machine

noplay:
The test server doesn't seem to freeze during the test because it's answer to the browser during the test session.

noplay:
I don't know if the trouble is specific to the way we do things into tests (it's a hack) or if it's hide a bigger issue.

noplay:
FINALLY I GOT IT

We need to use a dedicated TCP connector otherwise we conflict with the one of the server.

PR #1036, opened 2017-05-17
Aiohttp 2.0

noplay:
Ref #1000

### DIFF
--- a/gns3server/compute/docker/__init__.py
+++ b/gns3server/compute/docker/__init__.py
@@ -56,7 +56,7 @@ class Docker(BaseManager):
                 self._connected = True
                 connector = self.connector()
                 version = yield from self.query("GET", "version")
-            except (aiohttp.errors.ClientOSError, FileNotFoundError):
+            except (aiohttp.ClientOSError, FileNotFoundError):
                 self._connected = False
                 raise DockerError("Can't connect to docker daemon")
             if parse_version(version["ApiVersion"]) < parse_version(DOCKER_MINIMUM_API_VERSION):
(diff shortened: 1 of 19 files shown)


grossmj:
There are issues loading labs.

svg_issue

```
QSvgHandler: Width or height for "data:image/png;base64," image was not greater than 0
Cannot read file '/tmp/tmpxymtx2rq/b4625a5a801e68d455bd5cf0012a088f.svg', because: Premature end of document. (line 1)
QSvgHandler: Width or height for "data:image/png;base64," image was not greater than 0
```

noplay:
It's standard router image?

grossmj:
Yes

noplay:
It's with last 2.1 code for GUI? I didn't reproduce it :(

noplay:
But it's possible we introduce a bug with old Qt because the way to send file has changed now it's chunked I will try to make it with a content length ...

noplay:
Could you test again?

grossmj:
Seems that the symbol don't vanish anymore however I get "project not opened" messages all the time when moving a node on the scene...

project_not_opened

noplay:
**** buggy old Qt

noplay:
Which version of Qt are you using? 

I don't see anything wrong on HTTP level :( The response are the same and it doesn't fail on my computer.

grossmj:
I tested with latest versions: Python 3.5.2 Qt 5.8.0 and PyQt 5.8.2.

grossmj:
Same issue with Python 3.5.2 Qt 5.5.1 and PyQt 5.5.1.

grossmj:
Creating a lab from scratch, same problem :(

grossmj:
This was due to an old aiohttp version... :/

PR #1061, opened 2017-06-07
Fix when config file doesn't have computes section

ziajka:
Small fix, but it goes from 2.1 branch. Would you like to have from master?


noplay:
Can you open the PR against master?

noplay:
Closed we will reopen it against master

PR #1075, opened 2017-06-15
Fix the error about appliance ID not found

noplay:


noplay:
Could you test again?

grossmj:
It works for me. I think you can merge it.

PR #1091, opened 2017-06-27
Bugfix 557

ziajka:
Possible solution for the IOU problem.

### DIFF
--- a/gns3server/schemas/iou.py
+++ b/gns3server/schemas/iou.py
@@ -86,6 +86,10 @@ IOU_CREATE_SCHEMA = {
             "description": "Private-config of IOU",
             "type": ["string", "null"]
         },
+        "application_id": {
+            "description": "Application ID for running IOU image",
+            "type": ["integer", "null"]
+        },
     },
     "additionalProperties": False,
     "required": ["name", "path"]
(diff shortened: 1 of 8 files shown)


gns3server/schemas/iou.py
```diff
@@ -182,7 +186,11 @@
         "command_line": {
             "description": "Last command line used by GNS3 to start QEMU",
             "type": "string"
-        }
+        },
+        "application_id": {
+            "description": "Application ID for running IOU image",
+            "type": ["integer", "null"]
```
noplay: I think application_id in  IOU_OBJECT_SCHEMA  can't be null

tests/controller/test_project.py
```diff
@@ -383,7 +361,7 @@ def test_deleteDrawing(async_run, project, controller):
     assert len(project._drawings) == 0
 
 
-def test_cleanPictures(async_run, project, controller):
+def test_clean_pcictures(async_run, project, controller):
```
noplay: Typo for pictures

gns3server/schemas/iou.py
ziajka: We support old behavior (local generation) still, but of course we can force to define application_id.

noplay: If I'm correct the OBJECT is not use for creation but just when we return the info. In this case the application_id should have been generated

ziajka: Good point! Thank you.

PR #1092, opened 2017-06-27
Fix the initial startup for VPCS is not used

noplay:
The default VPCS template didn't provide the base
startup config setting. This prevent the replacement
of the hostname in the configuration file.

Fix https://github.com/GNS3/gns3-gui/issues/2123


grossmj:
Looks like there is a still a bug. The initial config is only generated for the first device I drop on the scene.

vpcs_bug

noplay:
Wow good catch.  This is so weird. If we restart the server the next node
will have a base config.

noplay:
Fixed by using deepcopy instead of just copy

PR #1103, opened 2017-07-04
Graphics settings

ziajka:
Reference: https://github.com/GNS3/gns3-gui/issues/2124
and
https://github.com/GNS3/gns3-gui/issues/712


noplay:
Sound good

PR #1105, opened 2017-07-06
Bugfix 2136 (gui)

ziajka:
Reference: https://github.com/GNS3/gns3-gui/issues/2136


noplay:
Sound good, I hope it will not break when we will merge it to 2.1 due to the change of aiohttp version

Issue #1115, opened 2017-07-10
Drop use_ubridge flag 


noplay:
We have a flag use_ubridge that allow to disable ubridge.

It's add complexity in the code in order to support with and without. And we don't test without ubridge. Some feature like ipv6, packet capture and packet filter work only with ubridge. Their is today no reason for disabling ubridge. 

I propose to simplify and always consider that ubridge is enable.

grossmj:
Yes we can probably drop it now.

noplay:
I think it's good thing we remove it. I just discover we don't test qemu with ubridge enable and the test fail

PR #1117, opened 2017-07-11
Remove use_ubridge flag

noplay:
Ref #1115

The PR is bigger than expected, because test was not running with use_ubridge enable

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -212,6 +212,7 @@ def run_around_tests(monkeypatch, port_manager, controller, config):
     config.set("VirtualBox", "vboxmanage_path", tmppath)
     config.set("VPCS", "vpcs_path", tmppath)
     config.set("VMware", "vmrun_path", tmppath)
+    config.set("Dynamips", "dynamips_path", tmppath)
 
     # Force turn off KVM because it's not available on CI
     config.set("Qemu", "enable_kvm", False)
--- a/tests/controller/test_project_open.py
+++ b/tests/controller/test_project_open.py
@@ -105,7 +105,6 @@ def demo_topology():
                     "node_id": "64ba8408-afbf-4b66-9cdd-1fd854427478",
                     "node_type": "vpcs",
                     "properties": {
-                        "startup_script": "",
                     },
                     "symbol": ":/symbols/computer.svg",
                     "width": 65,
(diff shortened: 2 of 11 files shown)


tests/controller/test_project_open.py
```diff
@@ -105,7 +105,6 @@ def demo_topology():
                     "node_id": "64ba8408-afbf-4b66-9cdd-1fd854427478",
                     "node_type": "vpcs",
                     "properties": {
-                        "startup_script": "",
```
noplay: Not related to the PR

tests/handlers/api/compute/test_vpcs.py
```diff
@@ -83,28 +84,19 @@ def test_vpcs_nio_update_udp(http_compute, vm):
                                     "rhost": "127.0.0.1",
                                     "filters": {}},
                                 example=True)
-    assert response.status == 201
+    assert response.status == 201, response.body.decode("utf-8")
```
noplay: Adding the response body allow when the test fail to see the content of the body to debug

tests/conftest.py
```diff
@@ -212,6 +212,7 @@ def run_around_tests(monkeypatch, port_manager, controller, config):
     config.set("VirtualBox", "vboxmanage_path", tmppath)
     config.set("VPCS", "vpcs_path", tmppath)
     config.set("VMware", "vmrun_path", tmppath)
+    config.set("Dynamips", "dynamips_path", tmppath)
```
noplay: Not related to the PR. Just prevent dynamips binary to be called from the test

grossmj:
I get an error with this PR. Steps to reproduce:

* Drop 2 VPCS nodes.
* Start the first one.
* Add a link between the 2 nodes.
* Start the second one.

vpcs_error

grossmj:
Same issue with 2 Qemu nodes.

grossmj:
Linking 2 VirtualBox VMs together: ```Error while creating link: Adapter 0 doesn't exist on VirtualBox VM 'LinuxTinycore3.8.2X-1'```

grossmj:
I am concerned this PR breaks something we won't notice. Should we postpone it after 2.1?

noplay:
I think we need to understand why it's fail. Because with filter we introduce more ubridge usage we need to make sure the code around ubridge is reliable.

noplay:
I think I got the problem. Half from the PR, half existing buggy code (VirtualBox).

We have two variable ubridge and use_ubridge and sometimes we have confusion between them.

noplay:
Sound good now

PR #1119, opened 2017-07-11
Test if a snapshot name already exists

### LINKED ISSUE #1118: Random test failure
Sometimes this test fail on our computers or on Travis. It seem to fail more when the computer is slow.

```
________________________________ test_snapshot _________________________________
project = <gns3server.controller.Project Test 64198a01-f818-4a5c-9772-a540a17616cf>
async_run = <function async_run.<locals>.<lambda> at 0x7f7b9713a9d8>
    def test_snapshot(project, async_run):
        """
        Create a snapshot
        """
        assert len(project.snapshots) == 0
    
        snapshot = async_run(project.snapshot("test1"))
        assert snapshot.name == "test1"
    
        assert len(project.snapshots) == 1
        assert list(project.snapshots.values())[0].name == "test1"
    
        # Raise a conflict if name is already use
        with pytest.raises(aiohttp.web_exceptions.HTTPConflict):
>           snapshot = async_run(project.snapshot("test1"))
E           Failed: DID NOT RAISE <class 'aiohttp.web_exceptions.HTTPConflict'>
tests/controller/test_project.py:537: Failed
```

noplay:
This fix random test failure when testing snapshots. It
seem under high load sometimes the previous snapshot folder
was not visible on disk. Perhaps a test isolation issue but
I don't see how.

But in any case it's better to test if the name is not already
use.

Fix #1118

### DIFF
--- a/gns3server/controller/project.py
+++ b/gns3server/controller/project.py
@@ -510,6 +510,9 @@ class Project:
         :param name: Name of the snapshot
         """
 
+        if name in [snap.name for snap in self.snapshots.values()]:
+            raise aiohttp.web_exceptions.HTTPConflict(text="The snapshot {} already exist".format(name))
+
         snapshot = Snapshot(self, name=name)
         try:
             if os.path.exists(snapshot.path):


noplay:
What kind of additional test can I add?

ziajka:
You're totally right, it's already tested. Haven't seen the context, I just took a look on commit.

noplay:
:D

PR #1125, opened 2017-07-12
Filter support for Docker

### LINKED ISSUE #1110: Filters support for Docker

noplay:
Fix #1110


gns3server/compute/docker/docker_vm.py
```diff
@@ -712,6 +722,19 @@ def adapter_add_nio_binding(self, adapter_number, nio):
...
+        :param adapter_number: adapter number
+        :param nio: NIO instance to add to the adapter
+        """
+
+        bridge_name = 'bridge{}'.format(adapter_number)
+        if bridge_name in self._bridges:
+            yield from self._ubridge_apply_filters(bridge_name, nio.filters)
+
+    @asyncio.coroutine
     def adapter_remove_nio_binding(self, adapter_number):
```
grossmj: Shouldn't you remove bridge_name from self._bridges in that method?

noplay: When the bridge is created we need to remember forever that we can apply filter on it.

The docker code, stop and start the bridge when the nio is changed but doesn't destroy the bridge.

PR #1130, opened 2017-07-18
Filters support for VMware

### LINKED ISSUE #1108: Filters support for VMware

noplay:
Fix #1108

### DIFF
--- a/gns3server/compute/vmware/vmware_vm.py
+++ b/gns3server/compute/vmware/vmware_vm.py
@@ -68,6 +68,10 @@ class VMwareVM(BaseNode):
         if not os.path.exists(vmx_path):
             raise VMwareError('VMware VM "{name}" [{id}]: could not find VMX file "{vmx_path}"'.format(name=name, id=node_id, vmx_path=vmx_path))
 
+    @property
+    def ethernet_adapters(self):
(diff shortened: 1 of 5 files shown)


grossmj:
I cannot ping between 2 VMware VMs as soon as a filter is applied.

grossmj:
Connecting to uBridge and adding a filter manually works.

grossmj:
I suspect some other problem at play, I can see the filter is applied if I connect manually to uBridge.

```
bridge show ethernet0.vnet
101 bridge 'ethernet0.vnet' is running
101 Filter 'filter0' configured in position 1
101 Source NIO: vmnet3
101 Destination NIO: 10000:127.0.0.1:10001
100-OK
```

However, something happens in the VM and a new IP address is set on the network adapter.

filter_applied

noplay:
Perhaps VMware decide to reconfigure the adapter for whatever reason

grossmj:
Yes because I cannot reproduce it if I apply a filter manually. Maybe the VMX file gets rewritten which trigger something in the VM? I am still investigating...

grossmj:
I am going to merge this since the problem doesn't seem to be related on the packet filter support itself.

PR #1138, opened 2017-07-25
Duplicate support for qemu

noplay:
Ref https://github.com/GNS3/gns3-gui/issues/1065


noplay:
Doesn't seem to require  more special code.

Issue #1083, opened 2017-06-22
History for the Ethernet Switch console


noplay:
The console of the ethernet switch (EmbedShell class) doesn't support history with the arrow key. This could be a good improvement especially because we will also use it in the embed VPCS.

PR #1139, opened 2017-07-25
Prompt with new vpcs, fixes: #1083

ziajka:


ziajka:
Seems to fail on Windows. Let me check it first.

requirements.txt
```diff
@@ -7,3 +7,4 @@ raven>=5.23.0
 psutil>=3.0.0
 zipstream>=1.1.4
 typing>=3.5.3.0 # Otherwise yarl fail with python 3.4
+prompt-toolkit
```
noplay: Do we need a minimal version?

gns3server/utils/asyncio/embed_shell.py
```diff
@@ -111,29 +127,130 @@ def run(self):
...
+        self._loop = loop
+        self._cli = None
+        self.encoding = 'utf-8'
+
+    @asyncio.coroutine
+    def connected(self):
+        self._cli = CommandLineInterface(
+            application=create_prompt_application(self._shell.prompt),
+            eventloop=UnstoppableEventLoop(create_asyncio_eventloop(self._loop)),
+            output=Vt100_Output(self, lambda: Size(rows=40, columns=79)))
```
noplay: It's bug if the terminal is smaller. I think we can get the size via telnet

requirements.txt
ziajka: I don't think so, actually cannot see any backlog of versions (to choose the minimal).

ziajka:
@noplay done :)

noplay:
Bravo it's work !!!

PR #1153, opened 2017-08-02
Allow use some API when project is close

noplay:
* /nodes
* /drawings
* /links

Fix https://github.com/GNS3/gns3-gui/issues/2189


noplay:
I'm not really confortable with merging it now in 2.1. Because their more side effect than expected and an error could lead to project corruption.

grossmj:
This needs to be based on 2.2 branch

Issue #1159, opened 2017-08-10
KeyError: 'link_id'


ziajka:
https://sentry.io/gns3/gns3-server/issues/324938488/

```
KeyError: 'link_id'
(1 additional frame(s) were not displayed)
...
  File "./gns3-server\gns3server\handlers\api\controller\project_handler.py", line 173, in load
  File "./gns3-server\gns3server\controller\__init__.py", line 551, in load_project
  File "./gns3-server\gns3server\utils\asyncio\__init__.py", line 154, in new_function
  File "./gns3-server\gns3server\controller\project.py", line 779, in open
  File "./gns3-server\gns3server\controller\project.py", line 751, in open

KeyError: 'link_id'
```

PR #1161, opened 2017-08-10
Fixes loading project when link_id is not set (#1159)

ziajka:
@grossmj I have doubts about this solution. I cannot judge if link_id is not set by user interaction in file (the most likely) or this bug has deeper source.


grossmj:
I cannot tell too, link_id is supposed to always be present. Maybe (I hope) this is some user playing with the file. We should have a schema checking if a project file is valid.

grossmj:
I cannot tell too, link_id is supposed to always be present. Maybe (I hope) this is some user playing with the file. We should have a schema checking if a project file is valid.

PR #1195, opened 2017-10-14
QEMU: fix logging of base mac address when creating a new node

ddragic:
gns3 server log doesn't show the base MAC address when creating a new qemu node (i.e when the mac address is auto-generated).


grossmj:
Thanks!

PR #1228, opened 2017-11-20
Close telnet console after stopping the node. Ref gui/#2330

ziajka:


grossmj:
Not instantaneous but works :+1:

PR #1252, opened 2018-01-05
GNS3-API: implement GET for specific drawing and link

### LINKED ISSUE #1249: GNS3-API: GET method not implemented for specific link or drawing
While it's possible to get the data of a specific node or project, it's not possible for links and drawings.

For the API endpoints [/v2/projects/{project_id}/links/{link_id}](http://api.gns3.net/en/latest/api/v2/controller/link/projectsprojectidlinkslinkid.html) and [/v2/projects/{project_id}/drawings/{drawing_id}](http://api.gns3.net/en/latest/api/v2/controller/drawing/projectsprojectiddrawingsdrawingid.html) no GET method is defined, only PUT and DELETE. Workaround is to get all links / drawing. Especially for drawings this can be quite inefficient as drawing can be large and having to load unneeded ones leads to a big overhead.

ehlers:
Fixes #1249

Initially I also wanted to include the changed `docs/api` files in this PR. But during the preparation I noticed, that lots of the current API documentation is outdated. Updating only a small part of the API documentation makes no sense, all the documentation should be updated. I'll create an issue for that.

### DIFF
--- a/gns3server/handlers/api/controller/drawing_handler.py
+++ b/gns3server/handlers/api/controller/drawing_handler.py
@@ -61,6 +61,26 @@ class DrawingHandler:
         response.set_status(201)
         response.json(drawing)
 
+    @Route.get(
+        r"/projects/{project_id}/drawings/{drawing_id}",
+        parameters={
+            "project_id": "Project UUID",
+            "drawing_id": "Drawing UUID"
+        },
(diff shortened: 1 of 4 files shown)


grossmj:
Thanks a lot :+1:

PR #1263, opened 2018-01-14
Rename ethernet switch shell command arp to mac

ddragic:
From https://gns3.com/qa/ethernet-switch-incorrectly-has-


grossmj:
Thanks a lot :+1:

Issue #1186, opened 2017-10-02
Deep dive into a way of storing configuration and resources (images, etc)


ziajka:
We need to revise configuration and sharing resources between user, controller and computes. Trace all route of storing files and think how we can improve it and solve bottlenecks. Probably unify a way of accessing configuration and separation between controller/computes.

Example of not consistent deletion:

https://github.com/GNS3/gns3-server/blob/2.1/gns3server/controller/project.py#L703

Subject is also related to importing/exporting projects. 

Ref. https://github.com/GNS3/gns3-gui/issues/2271

ziajka:
Ref. https://github.com/GNS3/gns3-gui/issues/2239

PR #1275, opened 2018-01-29
 Calculate MD5 on thread and before json response, Ref. gui#2239 

ziajka:
This should basically don't block GUI during computations. Further work requires change the way how we're taking care of images. In the long term we will look on that while working on computes changes (https://github.com/GNS3/gns3-server/issues/1186).


grossmj:
Beautiful :+1:

PR #1295, opened 2018-02-13
Implement a minimum interval between psutil calls. Fixes #2262

ehlers:
Ensures a minimum interval between two psutil.cpu_percent() calls. If CPU and memory are requested faster than this interval, the latest result is returned.


grossmj:
Thanks a lot, this is very helpful :)

PR #1302, opened 2018-02-20
Fix missing 'locales' package in base image

nabilbendafi:
`locales` package is missing from the Ubuntu 16.04 base image, thus building the gns3-server Docker image fails with following error message
 
```bash
/bin/sh: 1: locale-gen: not found
```


ziajka:
Thank you @nabilbendafi!

PR #1316, opened 2018-03-14
update Cisco ASAv appliance to include version 9.9.1

BrentBaccala:


grossmj:
Thanks! :+1:

ehlers:
While this contribution is very welcome. I think it's totally misplaced here. It changes the appliances in the master branch of the GNS33-server. But that's pretty useless, as this branch is not used to create any release. The changes will not be automatically merged to the 2.1 and 2.2 branch and the gns3-registry repository.

This correct way would be to add this to the gns3-registry, which is synced to this gns3-server repository just before a release.

Issue #851, opened 2016-12-31
Sporadically systemd is unable to start gns3-server


ehlers:
Self made GNS3VM with GNS3 Version 1.5 and 2.0 on Debian Testing

Since several months systemd sporadically can't start the gns3-server. The gns3-server is started, but remains in the status starting (it doesn't reach status running). After 90 seconds systemd kills the gns3-server.

I don't recall, that I had this issue in the beginning, so I assume it was introduced with a newer systemd or kernel version.

This happens only on cold start of the GNS3VM and the probability is quite high when the host OS (OS X in my case) is also cold started as well.

Here the systemd status of a successful start:
```
gns3@gns3vm:~$ sudo service gns3 status
● gns3.service - GNS3 server
   Loaded: loaded (/lib/systemd/system/gns3.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2016-12-31 09:44:35 CET; 1min 14s ago
  Process: 653 ExecStart=/usr/local/bin/gns3server --log /var/log/gns3/gns3server.log --pid /var/run/gns3/gns3server.pid --daemon (code=exited, status=0/SUCCESS)
  Process: 649 ExecStartPre=/bin/chown -R gns3:gns3 /var/log/gns3 /var/run/gns3 (code=exited, status=0/SUCCESS)
  Process: 646 ExecStartPre=/bin/mkdir -p /var/log/gns3 /var/run/gns3 (code=exited, status=0/SUCCESS)
 Main PID: 679 (gns3server)
    Tasks: 1 (limit: 19660)
   Memory: 52.4M
      CPU: 1.901s
   CGroup: /system.slice/gns3.service
           └─679 /usr/bin/python3 /usr/local/bin/gns3server --log /var/log/gns3/gns3server.log --pid /var/run/gns3/gns3server.pid --daemon

Dec 31 09:44:21 gns3vm systemd[1]: Starting GNS3 server...
Dec 31 09:44:30 gns3vm systemd[1]: gns3.service: PID file /var/run/gns3/gns3server.pid not readable (yet?) after start: No such file or directory
Dec 31 09:44:35 gns3vm systemd[1]: Started GNS3 server.
```

and here the status of a failed start:
```
gns3@gns3vm:~$ sudo service gns3 status
[sudo] password for gns3:
● gns3.service - GNS3 server
   Loaded: loaded (/lib/systemd/system/gns3.service; enabled; vendor preset: enabled)
[...]

PR #1322, opened 2018-04-03
Update remote-install.sh

tgupta3:
Updated systemd config

it seems that the install script tends to use a different config when the ubuntu distribution is not trusty. In my case, I observed gns3 was installed in /usr/local/bin but the config tends to look up in /usr/bin/gns3server. Additionally https://github.com/GNS3/gns3-server/issues/851 this issue would also pop up due to early start of the gns3 service in the boot process

### DIFF
--- a/scripts/remote-install.sh
+++ b/scripts/remote-install.sh
@@ -259,6 +259,8 @@ else
     cat <<EOFI > /lib/systemd/system/gns3.service
 [Unit]
 Description=GNS3 server
+Wants=network-online.target
+After=network.target network-online.target
 
 [Service]
 Type=forking
@@ -267,8 +269,8 @@ Group=gns3
 PermissionsStartOnly=true
 ExecStartPre=/bin/mkdir -p /var/log/gns3 /var/run/gns3
 ExecStartPre=/bin/chown -R gns3:gns3 /var/log/gns3 /var/run/gns3
-ExecStart=/usr/bin/gns3server --log /var/log/gns3/gns3.log \
-     --pid /var/run/gns3/gns3.pid --daemon
+ExecStart=/usr/local/bin/gns3server --log /var/log/gns3/gns3.log \
+ --pid /var/run/gns3/gns3.pid --daemon
 Restart=on-abort
 PIDFile=/var/run/gns3/gns3.pid
 


grossmj:
@tgupta3 thank you for your contribution! :+1: what distribution have you tested?

tgupta3:
Xenial 16.04

grossmj:
Something isn't right. The default install location for the GNS3 server should be `/usr/bin/gns3server`, at least that's what we have for our .deb packages (https://launchpad.net/~gns3/+archive/ubuntu/ppa). How did you install GNS3?

tgupta3:
Interesting, I didn't use the PPA to install it, since I only wanted to deploy it a remote server. So I used this script https://github.com/GNS3/gns3-server/blob/master/scripts/remote-install.sh by following the instructions from http://docs.gns3.com/1c2Iyiczy6efnv-TS_4Hc7p11gn03-ytz9ukgwFfckDk/

Moreover, when I needed to update I did by using pip3
pip3 install gns3-server -U

grossmj:
The script install the server using apt-get: https://github.com/GNS3/gns3-server/blob/master/scripts/remote-install.sh#L158

Maybe the script could check both location?

tgupta3:
Let me try it on a fresh ubuntu 14 as well as Ubuntu 16, and see what's the behavior

tgupta3:
Did some additional tests today. Seems like for both ubuntu 14.04 and Ubuntu 16.04 GNS3 server is being installed in /usr/bin/gns3server. However, on updating gns3-server using pip3, I found it's installing the entry point script in /usr/local/bin

candlerb:
> Did some additional tests today. Seems like for both ubuntu 14.04 and Ubuntu 16.04 GNS3 server is being installed in /usr/bin/gns3server. However, on updating gns3-server using pip3, I found it's installing the entry point script in /usr/local/bin

Those are two different installation methods - `apt-get install` and `pip3 install` - and they install to different locations.  If you install using one, you should upgrade using the same one.

The `remote-install.sh` script you linked to uses `apt-get install` and the systemd file it installs points to the correct location for that method.

So I'd say everything is correct as it is today.  If you install using `pip3 install` then creating a suitable systemd unit file is up to you.

PR #1348, opened 2018-06-07
Update API documentation

ehlers:
As some API calls have been extended since the last documentation update on v2.1.2, I think that it's time for another update.

As these updates are very big, it's hard to see if anything is missing. But I checked, if bigger parts from the previous version were deleted (what's not the case) and that the latest changes are included (nodes/links, project/variables).


grossmj:
Thanks a lot :+1:

PR #1351, opened 2018-06-11
Don't release NIO UDP ports when updating docker container.

acaudwell:
When a docker container was updated, the NIO UDP ports were being released
from port_manager, but they were still stored in self._ethernet_adapters.
When the container was restarted, the same NIO ports would be used. However,
if a new connection was then added, it could assign a port that was already
in use. The end result was that traffic would be bridged across the wrong
interface connection.


grossmj:
Thanks for catching that! :+1:

PR #1368, opened 2018-07-12
Require typing only for Python < 3.5

sbraz:
Hi,
The typing package is already included in Python 3.5 and later and it can apparently cause failures for Python 3.7.


grossmj:
Thanks! :+1:

PR #1391, opened 2018-08-20
 Allow virtual machines to use files in project directory as disk images

BrentBaccala:
This is useful for Cisco VMs that will load a boot configuration if it appears on an ISO filesystem attached as a virtual CD-ROM, and boot configurations are logically part of the project, not global.


grossmj:
I have created a new PR based on yours: https://github.com/GNS3/gns3-server/pull/1462 I will merge it soon. Thanks for your contribution! :+1:

PR #1476, opened 2018-11-28
Docker: fix x11vnc not starting

ddragic:
x11vnc fails to start because of a typo:

vnc.log:
29/11/2018 00:24:06 *** unrecognized option(s) ***
29/11/2018 00:24:06 	[1]  -nopw-shared


grossmj:
Thanks for catching this :+1:

PR #1572, opened 2019-04-05
Support for docker images that set the USER directive

kazkansouh:
Some docker images explicitly set the [runtime user](https://docs.docker.com/engine/reference/builder/#user) such as [graylog](https://hub.docker.com/r/graylog/graylog/). This means that the `init.sh` script used by gns3 inside docker will not work as it need root permissions.

This patch starts the container with root user, then uses `su` to drop to the desired user (or root if no user was configured) after `init.sh` has finished.

The `init.sh` has a little additional complexity to handle nested arguments, such as the convoluted example:
```
            "Cmd": [
                "/bin/sh",
                "-c",
                "cd /etc ; pwd ; exec /bin/sh -c \"cd /etc/network ; pwd ; exec /bin/sh\""
            ],
```
Please feel free to update if there is a better way to handle these cases.

### DIFF
--- a/gns3server/compute/docker/docker_vm.py
+++ b/gns3server/compute/docker/docker_vm.py
@@ -311,6 +311,7 @@ class DockerVM(BaseNode):
             "Tty": True,
             "OpenStdin": True,
             "StdinOnce": False,
+            "User": "root",
             "HostConfig": {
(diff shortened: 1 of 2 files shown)


kazkansouh:
Failing tests, ill update test cases and recreate pull request.

PR #1573, opened 2019-04-05
Support for docker images that set the USER directive (+test coverage)

kazkansouh:
(updated pull request https://github.com/GNS3/gns3-server/pull/1572 with an additional commit to fix failing tests)

Some docker images explicitly set the [runtime user](https://docs.docker.com/engine/reference/builder/#user) such as [graylog](https://hub.docker.com/r/graylog/graylog/). This means that the `init.sh` script used by gns3 inside docker will not work as it need root permissions.

This patch starts the container with root user, then uses `su` to drop to the desired user (or root if no user was configured) after `init.sh` has finished.

The `init.sh` has a little additional complexity to handle nested arguments, such as the convoluted example:
```
            "Cmd": [
                "/bin/sh",
                "-c",
                "cd /etc ; pwd ; exec /bin/sh -c \"cd /etc/network ; pwd ; exec /bin/sh\""
            ],
```
Please feel free to update if there is a better way to handle these cases.

### DIFF
--- a/gns3server/compute/docker/docker_vm.py
+++ b/gns3server/compute/docker/docker_vm.py
@@ -311,7 +311,6 @@ class DockerVM(BaseNode):
             "Tty": True,
             "OpenStdin": True,
             "StdinOnce": False,
-            "User": "root",
             "HostConfig": {
                 "CapAdd": ["ALL"],
                 "Privileged": True,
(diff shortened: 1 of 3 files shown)


gns3server/compute/docker/docker_vm.py
```diff
@@ -342,6 +342,11 @@ def _create_network_config(self):
         # Give the information to the container the list of volume path mounted
         params["Env"].append("GNS3_VOLUMES={}".format(":".join(self._volumes)))
 
+        # Pass user configured for image to init script
+        if image_infos.get("Config", {"User": ""}).get("User"):
+            params["User"] = "root"
```
grossmj: Just a question. Why is root needed here?

kazkansouh: This tells docker to start the user inside docker container as root, so when the "init.sh" script runs it can configure the network. This is normally the default.

However, some images that are hardened for production use explicitly change the user inside the docker image to a non-root user which prevents the init.sh script from configuring the interfaces and hence does not work in GNS3.

grossmj:
Thanks for your contribution :)

grossmj:
There is an issue with Travis. I successfully ran the test locally

Issue #1595, opened 2019-06-03
adosztal/net-toolbox docker error on GNS server 2.2


32e3e32:
Hi all,
First a big thank you to everyone contributing to GNS3!!!

Now to the problem: I'm getting an error when adding docker device to topology. GNS Server 2.2.0dev12 running on Ubuntu 18. Later reinstalled Ubuntu and updated server and client to 2.2.0b2 but problem with "duplicate persistent volumes" is also "persisting"...

FYI, Other docker containers like webterm are working fine.

Here's the log:
`
2019-06-03 02:36:50 INFO web_log.py:233 192.168.0.211 [03/Jun/2019:02:36:50 +0000] "DELETE /v2/templates/51ca3751-6d1c-4069-af49-ed9e758f95d8 HTTP/1.1" 204 213 "-" "GNS3 QT Client **v2.2.0dev12**"
2019-06-03 02:37:16 INFO notification_handler.py:82 Client has disconnected from controller WebSocket
2019-06-03 02:37:16 INFO project_handler.py:279 Client has disconnected from notification stream for project ID '2040f60a-7ffa-4263-b3df-ae126169ce27' (WebSocket method)
2019-06-03 02:37:21 INFO project_handler.py:288 Project '2040f60a-7ffa-4263-b3df-ae126169ce27' is automatically closing due to no client listening
2019-06-03 02:37:21 INFO web_log.py:233 127.0.0.1 [03/Jun/2019:02:37:21 +0000] "POST /v2/compute/projects/2040f60a-7ffa-4263-b3df-ae126169ce27/ethernet_switch/nodes/1158f330-e9f8-43c5-9f3a-8a573fa2943a/stop HTTP/1.1" 204 256 "-" "Python/3.6 aiohttp/3.5.4"
2019-06-03 02:37:21 INFO ethernet_switch.py:248 Ethernet switch "Ethernetswitch-1" [1158f330-e9f8-43c5-9f3a-8a573fa2943a] has been deleted
2019-06-03 02:37:21 INFO hypervisor.py:148 Stopping Dynamips process PID=2399
2019-06-03 02:37:21 INFO project.py:361 Project 2040f60a-7ffa-4263-b3df-ae126169ce27 with path '/home/gns3/GNS3/projects/2040f60a-7ffa-4263-b3df-ae126169ce27' closed
[...]

PR #1597, opened 2019-06-05
Fix for issue #1595

kazkansouh:
Possible fix for https://github.com/GNS3/gns3-server/issues/1595

* Removes "Duplicate persistent volume" error message,
* Handles "/etc/network" mount consistently, and
* Selects most generic mount.


grossmj:
Awesome, I tested and everything seems to work well. Thanks a lot! :+1:

PR #1648, opened 2019-09-03
Add id value to all qemu drives

neffs:
A fixed id doesn't hurt and now we can select Disk Interface=none in
GNS3 and e.g. attach the volume as a USB thumb drive with -device
usb-storage,drive=drive0,...

It's also possible to override some options this way, such as the serial number of the ide drive.


neffs:
Can this still be added to a 2.2 release? This is only a minor change and would help a lot.

grossmj:
Thanks :+1:

PR #1649, opened 2019-09-03
Don't specify the PCI bus for AHCI device

neffs:
pci.X is not always defined, let qemu handle this.

I think PCI buses only exist if PCI network adapters are defined, this
is not the case with virtio.


neffs:
This is a bug in the current RC, so it probably should get fixed in 2.2. I noticed it when testing PR #1648.

PR #1691, opened 2019-11-10
Added workaround for #1690

### LINKED ISSUE #1690: --config option unexpected behaviour
When using the --config option (e.g `[gns3admin@desktop] ~: gns3server --config gns3_server.conf`), the behaviour for finding the "gns3_controller.conf" file is not working as expected.

If the following file layout exists:
```
/home/gns3admin/
└── GNS3
    ├── gns3_controller.conf
    ├── gns3_gui.conf
    └── gns3_server.conf
```
Running the following commands works as expected (server is started with ~/GNS3/gns3_controller.conf as controller file):
```
[gns3admin@desktop] ~: gns3server --config GNS3/gns3_server.conf
[gns3admin@desktop] ~/GNS3: gns3server --config ./gns3_server.conf
```

However, running the command
```
[gns3admin@desktop] ~/GNS3: gns3server --config gns3_server.conf
```
works, but the server sends a warning 
```
2019-11-10 19:25:53 ERROR __init__.py:169 Cannot write controller configuration file 'gns3_controller.conf': [Errno 2] No such file or directory: ''
```
and none of the controller settings are saved by the server.

I'm not sure of an "appropriate" solution, but I guess either adding a command-line option for the controller configuration or somehow making the directory default to `.` when it's not specified but the --config option is used might work.

JCDraaijer:
Fixes #1690

### DIFF
--- a/gns3server/config.py
+++ b/gns3server/config.py
@@ -47,7 +47,7 @@ class Config:
         if files and len(files):
             directory_name = os.path.dirname(files[0])
             if not directory_name or directory_name == "":
-                files[0] = './' + files[0]
+                files[0] = os.path.dirname(os.path.abspath(files[0])) + os.path.sep + files[0]
             self._main_config_file = files[0]
         else:
             self._main_config_file = None
--- a/gns3server/config.py
+++ b/gns3server/config.py
@@ -45,6 +45,9 @@ class Config:
         self._files = files
         self._profile = profile
         if files and len(files):
+            directory_name = os.path.dirname(files[0])
+            if not directory_name or directory_name == "":
+                files[0] = './' + files[0]
             self._main_config_file = files[0]
         else:
             self._main_config_file = None
--- a/.gitignore
+++ b/.gitignore
@@ -58,4 +58,5 @@ startup.vpcs
 
 # Virtualenv
 env
+venv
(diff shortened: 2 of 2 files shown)


gns3server/config.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, files=None, profile=None):
         self._files = files
         self._profile = profile
         if files and len(files):
+            directory_name = os.path.dirname(files[0])
+            if not directory_name or directory_name == "":
+                files[0] = './' + files[0]
```
grossmj: I would use files[0] = os.path.abspath(files[0]). This will make sure it works on Windows as well.

JCDraaijer: I may have misunderstood what files[0] is supposed to represent in this case. 

Since we're assigning  `self._main_config_file = files[0]`, `files[0]` (and by extension `self._main_config_file`) is supposed to point to the actual configuration file and not a directory right? Or is the behaviour different on Windows?

grossmj: Correct files[0] points to the actual configuration file, not a directory. Actually, I think both solution would work. I am just a bit concerned that the './' would not work on Windows.

grossmj:
Thanks :+1:

PR #1693, opened 2019-11-18
V2.1.21 zstdsnap - Replace zipstream with facebook zstandard compression

gleyfer:
I did not see https://github.com/GNS3/gns3-server/pull/1537 prior to completing this, but I wanted to at least share my work and see if it can be incorporated at some point.

I have added functions in the snapshot/project export/project import/project.py to utilize the python-zstandard library and its substantially faster multi-threaded encoder. The encoder achieves much better compression ratios with MUCH faster throughput. It also offers much more configuration to choose between speed or file size.

Rather than using a zipstream, I used a regular python dict to store the file paths/arcnames and pushed them into a tarfile which is feeding into a zstandard streaming compressor using their streaming API.

You can see the performance differences here: https://github.com/gleyfer/gns3-server/blob/v2.1.21-zstdsnap/README.rst

Snapshotting one of my source project folder at 154GB was taking 2+ hours, longer than the default TCP socket timeout on my server and it would error out on the GUI side. The produced snapshot was around 69GB . With the zstandard at compression level 11, long range matching enabled and a 2GB window, I am able to compress to a 38GB file from the same source project within 30 minutes.

The default compression settings in my commits are very aggressive and assume a powerful CPU/a lot of memory but produce a much better compression ratio at 4x+ the speed of zipstream. With less complicated compression strategies and faster backing storage it is possible to achieve much higher throughputs with much lower memory utilization.

Currently, I only have this set to be done for snapshots. For regular project export/import I use the normal zipstream. This is to maintain compatibility with normal GNS3 versions. It will also list and import the old snapshots as well using the existing code.

There are some obvious things I haven't gotten to yet (and many more I'm sure I am too much of a novice to know):

1. My dict isn't sorted/project.gns3 isn't at the very beginning of the snapshot file so I'm not verifying validity first then extracting later. It is the opposite, where I run an extractall and then check the project.gns3. This can be easily fixed.
2. I based all of my commits off 2.1.21 as this was the version I had installed and could test on.
3. This solution isn't currently portable. It requires both ends to be patched to import projects between different gns3server instances. Perhaps some gui option could be added to choose type. For now I just have it using the traditional zipstream for export/import.
4. No way to control the compression level from the GUI, needs to be manually adjusted in the python file at the moment.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,224 +1,70 @@
-GNS3-server
+GNS3-server-zstdsnap
 ===========
 
-.. image:: https://travis-ci.org/GNS3/gns3-server.svg?branch=master
-    :target: https://travis-ci.org/GNS3/gns3-server
+This is a branch off of the GNS3 2.1.21 release to support much faster snapshotting
+of large projects by replacing zipstream with the facebook zstandard streaming 
+compressor: https://github.com/facebook/zstd
 
-.. image:: https://img.shields.io/pypi/v/gns3-server.svg
-    :target: https://pypi.python.org/pypi/gns3-server
+The python-zstandard library is used for this purpose: 
+https://github.com/indygreg/python-zstandard
 
-This is the GNS3 server repository.
+The zstandard library provides multi-threaded compression and much more tunable 
+compression settings which provide much improved speed and compression ratios.
 
-The GNS3 server manages emulators such as Dynamips, VirtualBox or Qemu/KVM.
-Clients like the GNS3 GUI controls the server using a HTTP REST API.
+The zstandard compression will ONLY be applied to newly created *snapshots*. This branch
+has full interoperability with standard GNS3 snapshots and will just use the existing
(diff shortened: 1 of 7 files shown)


grossmj:
Thanks a lot for this contribution. It definitely looks great! I will review and will most likely implement this in version 2.3 or higher.

grossmj:
@gleyfer we are going to incorporate your work in version 3.0. I also heard about lz4 and it definitely look interesting, do you have an opinion about it?

cristian-ciobanu:
@grossmj Are the changes in pull request be ported for the GNS3 version 3.0 ?
I think this will provide a huge benefit for exporting very large projects with tens or hundreds of gigs in size. Currently the process is very slow.

grossmj:
@cristian-ciobanu I agree, the current import/export is very slow, especially for large projects. I will prioritize this for the 3.0 release.

PR #1721, opened 2020-03-02
Added Debian 10 minimal template, externally hosted

sabueso:
Hi

This is a simple and minimal Debian 10 appliance, AFAIK only Ubuntu distribution was added in GNS3.

Will be glad if you merge this PR into master, would be great to see Debian 10 as an option ;)

Thanks

### DIFF
--- a/gns3server/appliances/debian10-min.gns3a
+++ b/gns3server/appliances/debian10-min.gns3a
@@ -1,0 +1,44 @@
+{
+    "name": "Debian 10 Minimal",
+    "category": "guest",
+    "description": "Debian 10 Custom, with automatic disk resize and ssh/nmap",
+    "vendor_name": "Debian",
+    "vendor_url": "https://debian.org",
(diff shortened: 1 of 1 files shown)


b-ehlers:
This PR is misplaced here. Appliances should be added to the [GNS3 registry](https://github.com/GNS3/gns3-registry).

Some further comments:
- Currently the download website is not accessible. Some minutes ago the domain doesn't even exist, currently the HTTPS certificate is invalid
- This seems your private website, please be aware that hosting a 2.6GB large image can generate **a lot** of traffic.
- Why not using readily available images like https://www.osboxes.org/debian/ ?

sabueso:
@b-ehlers thanks for your comments

- The website was ontime provisioning, is up and running now
- It's a commercial website, no problem about traffic
- I'm not used with this kind of images (not particular reason, btw)

I will close this PR and will open it in the right place!

Cheers

PR #1735, opened 2020-04-06
Add node start delay in seconds

piotrpekala7:


gns3server/controller/project.py
```diff
@@ -1064,17 +1064,18 @@ def dump(self):
...
 
     @open_required
-    async def start_all(self):
+    async def start_all(self, delay = 0):
         """
         Start all nodes
         """
         pool = Pool(concurrency=3)
         for node in self.nodes.values():
+            await asyncio.sleep(delay)
```
piotrpekala7, line 1073: I see that we are using asyncio.sleep in pool.py#45 but I'm not sure if it's good practise. It it working in that case?

grossmj: I don't think this works. Probably the delay should be implemented in node.start()

ziajka: Yes, here https://github.com/GNS3/gns3-server/blob/master/gns3server/compute/qemu/qemu_vm.py#L942 would be good start

grossmj:
Implementation will be different on version 3.0

PR #1742, opened 2020-04-15
Add clipboard support for TigerVnc

fadlytabrani:
There's currently no clipboard support for servers running on TigerVnc. TigerVnc clipboard support is provided by vncconfig that needs to be spawned for every display socket. The code changes above provides that functionality


grossmj:
Thanks, we are going to review your contribution soon! :)

Issue #1762, opened 2020-05-23
Reinitializing node MAC addresses when duplicating projects causes issues.


fadlytabrani:
When duplicating projects, node mac addresses are reinitialized. Linux network devices such as Vyos that we use extensively detects the MAC change, instantiates a new interface name. This makes all configurations invalid. 

This behaviour was changed after the following issue was raised -> #1522 

To correct the previous issue, the fix should have been to reinitialise cloud nodes only instead of all nodes in the project.

grossmj:
> To correct the previous issue, the fix should have been to reinitialise cloud nodes only instead of all nodes in the project.

As commented on https://github.com/GNS3/gns3-server/pull/1763, this isn't that simple. One alternative would be to provide the option to reset MAC addresses when an user export or import a project.

For instance, VMware and VIrtualBox have the option to reset the MAC addresses when you import a VM as an OVA. This is probably the best solution.

fadlytabrani:
This option should be presented during duplication and exporting of projects. 

![image](https://user-images.githubusercontent.com/1628412/82726861-3e70f980-9d2a-11ea-99b5-cf96c1717459.png)

![image](https://user-images.githubusercontent.com/1628412/82726874-56e11400-9d2a-11ea-8ca5-0fd031fbf86c.png)

grossmj:
I have added the option for exporting and duplicating a project.

PR #1763, opened 2020-05-23
Update export_project.py. Limit reinitializing MAC addresses to cloud nodes.

fadlytabrani:
Limit reinitializing MAC addresses to cloud nodes. Resolves issue #1762 .

### DIFF
--- a/gns3server/controller/export_project.py
+++ b/gns3server/controller/export_project.py
@@ -193,9 +193,10 @@ async def _patch_project_file(project, path, zstream, include_images, keep_compu
                 if "properties" in node and node["node_type"] != "docker":
                     for prop, value in node["properties"].items():
 
-                        # reset the MAC address
-                        if reset_mac_addresses and prop in ("mac_addr", "mac_address"):
-                            node["properties"][prop] = None
+                        # reset the MAC address, limited to cloud nodes. 
+                        if node["node_type"] == "cloud":
+                            if reset_mac_addresses and prop in ("mac_addr", "mac_address"):
+                                node["properties"][prop] = None
 
                         if node["node_type"] == "iou":
                             if not prop == "path":


grossmj:
@fadlytabrani thanks for the PR however this won't work because "Cloud" nodes are more like a pseudo node to show an external connection in GNS3. A cloud node itself has no MAC address property.

Something like this would work:

1. For all nodes with the "mac_addr" or "mac_address" property.
2. Find all the interfaces is connected to a cloud.
3. Change/reset the mac address for this interface.

However, the MAC address property is at the node level, so you would have to reset the MAC address for all the interfaces. We do support custom interfaces for some node types but I believe we shouldn't add or change custom interfaces when an user export or import a project.

I will comment on https://github.com/GNS3/gns3-server/issues/1762 to provide an alternative.

fadlytabrani:
Agreed. I have just realized that after looking at the cloud node properties. Generally I disable cloud nodes and utilise nat instead. Can you consider the following options?


1. Is there anyway we can do a discovery via topology links and only reinitialize mac addresses for nodes/ports that are attached to cloud nodes?

2. Since resetting mac addresses is an option in the function, can we expose it as part of an option (checkbox) when doing exports in the GUI?

grossmj:
> Is there anyway we can do a discovery via topology links and only reinitialize mac addresses for nodes/ports that are attached to cloud nodes?

We could do something like that but this would add too much complexity.

> Since resetting mac addresses is an option in the function, can we expose it as part of an option (checkbox) when doing exports in the GUI?

This would be a better approach as explained in https://github.com/GNS3/gns3-server/issues/1762

I can take care of this, I just need to expose the option in either the export or import function.

grossmj:
This has been taken care in https://github.com/GNS3/gns3-server/pull/1764

PR #1781, opened 2020-06-16
Refactor tests

grossmj:
* Use pytest-aiohttp
* Use the async def / await syntax.
* Fix tests to run with Python 3.8


grossmj:
Recreated on correct branch: https://github.com/GNS3/gns3-server/pull/1782

PR #1793, opened 2020-07-01
adding freeRouter appliance

mc36:
please accept my small addition.... thanks!


grossmj:
This is not needed. The appliances will be imported from the registry repository before the next release.

PR #1801, opened 2020-07-19
QEMU config disk - enable QEMU config import/export

grossmj:


grossmj:
Should we have an option to activate/deactivate this? This could be a problem to add a disk behind users' back...

b-ehlers:
It's already there, but hidden/undocumented. The config disk uses the HD-D interface type, setting it to `none` disables it.

But I agree, that this config disk may be surprising for the users. Maybe an explicit option in the GUI might be better.

grossmj:
> It's already there, but hidden/undocumented. The config disk uses the HD-D interface type, setting it to none disables it.

Do you mean in the GUI or using the API directly?

b-ehlers:
I used the GNS3 GUI to change the disk interface of HD-D (Secondary Slave) of a QEMU VM. But as the API is used to transport these settings to the server, setting it via API should also work, but I haven't test his.

grossmj:
I have added the option (deactivated by default). Also, I explicitly add the config disk to hdd so it shows up in the GUI https://github.com/GNS3/gns3-server/pull/1801/commits/56aba96a5fb57f502b283a0bc543da415bb3943a#diff-7aedd6be975599a25247f67edca649e8R1842

Screenshot from 2020-08-14 18-04-03

I will merge this if everything is good on your side.

b-ehlers:
Thanks a lot for your changes.

I have only one comment:
I'm not sure, that the addition of `self.hdd_disk_image = disk` in <https://github.com/GNS3/gns3-server/commit/56aba96a5fb57f502b283a0bc543da415bb3943a#diff-7aedd6be975599a25247f67edca649e8R1842> is working as intended.

I haven't tested this, but it should work like this:

If self.hdd_disk_image is set, it is used as a normal linked base disk. The config disk is only used when self._create_config_disk is set and self.hdd_disk_image is not set. I fear, that the config disk is working only on the first start of a VM. Then the hdd_disk_image is set and the config disk feature is no longer used.

Have you tested that the config disk is working even when the hdd_disk_image  is shown?

Two additional suggestions:
- Maybe the HDD image name could be shown as unselectable in the GUI when the config disk option is set, so the user doesn't try to change the image name.
- Currently (commit https://github.com/GNS3/gns3-server/pull/1801/commits/b672900406e5ce10251cc1e231fee1d117ca3805) it uses the hdd_disk_interface only, when it's not "ide", otherwise it uses hda_disk_interface. The idea was, that I didn't want the config disk to be the first disk to be found by the kernel and be used as the boot disk. But that condition is a bit surprising for the user, maybe the fallback to the hda_disk_interface should go.

grossmj:
> If self.hdd_disk_image is set, it is used as a normal linked base disk. The config disk is only used when self._create_config_disk is set and self.hdd_disk_image is not set. I fear, that the config disk is working only on the first start of a VM. Then the hdd_disk_image is set and the config disk feature is no longer used.

I have decided to remove this. The original intent was to show users what we are doing in the back-end but now we have the checkbox in the GUI so that should be enough.

> Maybe the HDD image name could be shown as unselectable in the GUI when the config disk option is set, so the user doesn't try to change the image name.

Good idea. I have implemented this.

> Currently (commit b672900) it uses the hdd_disk_interface only, when it's not "ide", otherwise it uses hda_disk_interface. The idea was, that I didn't want the config disk to be the first disk to be found by the kernel and be used as the boot disk. But that condition is a bit surprising for the user, maybe the fallback to the hda_disk_interface should go.

I have tried to improve this part. Now all disk interfaces are set to "none" by default so they have to be explicitly configured (there is a fail-safe to fallback to "ide" if an image is set but not the interface). HDD will use the HDA disk interface if its own disk interface hasn't been explicitly set, I think this gives a chance to users to configure it if they want to.

b-ehlers:
Looks good, I think we are now ready to merge this.

I have two additional ideas for the GUI, but that could be done in the future:
- show the import/export/edit config option only, when the config option is selected for that VM
  That is tricky as currently `hasattr(instance, "exportConfigs")` is used to test if a config is available. My idea is to always have the exportConfigs method, but test if it returns an non-empty list. But as this is used in a lot of places this will be a large change.
- the edit config makes no sense for binary config data like the zip file. But I have no good idea how to limit the edit function to text configurations.

PR #1818, opened 2020-09-12
One GNS3 Server | Multiples GNS3 clients (client Isolation) 

ACGns3:
One GNS3 Server | Multiples GNS3 clients (client Isolation)
There is no user isolation, all users can view all projects and modify them.
Any plans to work  Isolating GNS3 clients  Connecting to the GNS3 Server and isolates Projects, settings, and avoid being modified by other clients .

### DIFF
--- a/gns3server/version.py
+++ b/gns3server/version.py
@@ -23,8 +23,8 @@
 # or negative for a release candidate or beta (after the base version
 # number has been incremented)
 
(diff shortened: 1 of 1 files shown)


grossmj:
We are working on user management and rights etc. We are expecting to have something ready by the end of the year or early next year.

PR #1837, opened 2020-11-03
fixed limit of 512 ious being across all computes, instead of per-com…

Vladilit:
Although the comment in this section says that there should be a limit of 512 ious **per compute**, there is currently a single global limit of 512 ioud for all computes together.
This is a simple fix for that, making the functionality work as intended - 512 ious per compute.


Vladilit:
Not sure why the auto-check is failing.. it works fine for me on 3.7

grossmj:
Thanks for the contribution. We had already fixed it on another branch. Please test with the latest version and let me know if it doesn't work as intended.

PR #1856, opened 2020-12-30
Allow cloned QEMU disk images to be resized before the node starts

BrentBaccala:
Clone the disk image in response to a resize request instead of waiting until the node starts.


grossmj:
Thanks :+1:

PR #1905, opened 2021-05-24
Preserve auto_start/auto_open/auto_close when restoring snapshot

### LINKED ISSUE #1904: Preserve "Leave this project running in the background after closing" when restoring a snapshot
After restoring a snapshot, the flag "Leave this project running in the background after closing" is cleared.  This is the case even if the `project.gns3` file in the snapshot has `"auto_close": false`.

AFAICS, the `restore` function in `controller/snapshot.py` calls `import_project`, which I think is the same as importing a portable project, but that resets the flags [here](https://github.com/GNS3/gns3-server/blob/master/gns3server/controller/import_project.py#L100-L103).

I think that if you're restoring a snapshot in an existing project, it would be better to keep these flags the same as they were previously in the project.

candlerb:
Fixes #1904

(Fix made and tested on 2.2 branch; let me know if you want this rebasing to master)

### DIFF
--- a/gns3server/controller/import_project.py
+++ b/gns3server/controller/import_project.py
@@ -38,7 +38,8 @@ Handle the import of project from a .gns3project
 """
 
 
-async def import_project(controller, project_id, stream, location=None, name=None, keep_compute_id=False):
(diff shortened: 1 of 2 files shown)


grossmj:
Thanks :+1:

PR #1908, opened 2021-06-03
Use a stock BusyBox for the Docker Integration

grossmj:


b-ehlers:
Just one small addition: Please add a short note to <https://github.com/GNS3/gns3-server/blob/3.0/README.rst>, that Docker support needs the script program (package bsdutils or util-linux), when running a docker VM and a static busybox during installation (`python3 setup.py install` / `pip3 install` / package creation).

PR #1910, opened 2021-06-04
Set py-cpuinfo to 8.0.0

fabaff:
`py-cpuinfo` was released a while back. According to the [Changelog](https://github.com/workhorsy/py-cpuinfo/blob/master/ChangeLog) it looks like a bugfix release.

The requirement are pinned and this causes build failures for distribution packages if the don't match the latest available release.


grossmj:
Thanks :+1:

PR #1927, opened 2021-06-30
Support cloning of encrypted qcow2 base image files

candlerb:
This is a minimal changeset to address #1921.

This patch parses the json output from `qemu_img info`, both as a way of validating the image without calling `rebase`, and to obtain the base image virtual-size and encryption flag.

It uses the existing `_qemu_img_exec`, which mixes stdout and stderr together and writes them to `_qemu_img_stdout_file`, and is then read back by `read_qemu_img_stdout()`.  It works because the stderr output happens to be empty when there's valid info output on stdout.

Arguably it would be cleaner to separate stdout and stderr, either to two separate files, or a stderr file and RAM, or just read them both into RAM: e.g.

```
process = await asyncio.create_subprocess_exec(*command, stdout=asyncio.subprocess.PIPE,
                        stderr=asyncio.subprocess.PIPE, cwd=self.working_dir)
(stdout_data, stderr_data) = await process.communicate()
log.info("{} returned with {}".format(self._get_qemu_img(), process.returncode))
return (process.returncode, stdout_data, stderr_data)
```

But that would be a more invasive change (and I'm not sure if the output file is wanted on disk for other purposes)

### DIFF
--- a/gns3server/compute/qemu/qemu_vm.py
+++ b/gns3server/compute/qemu/qemu_vm.py
@@ -1675,6 +1675,24 @@ class QemuVM(BaseNode):
         try:
             qemu_img_path = self._get_qemu_img()
             command = [qemu_img_path, "create", "-o", "backing_file={}".format(disk_image), "-f", "qcow2", disk]
+            try:
+                base_qcow2 = Qcow2(disk_image)
+                if base_qcow2.crypt_method:
+                    # Workaround for https://gitlab.com/qemu-project/qemu/-/issues/441
+                    # Also embed a secret name so it doesn't have to be passed to qemu -drive ...
+                    options = {
+                        "encrypt.key-secret": os.path.basename(disk_image),
+                        "driver": "qcow2",
+                        "file": {
+                            "driver": "file",
+                            "filename": disk_image,
+                        },
+                    }
+                    command = [qemu_img_path, "create", "-b", "json:"+json.dumps(options, separators=(',', ':')),
+                               "-f", "qcow2", "-u", disk, str(base_qcow2.size)]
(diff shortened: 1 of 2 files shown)


candlerb:
I noticed that `gns3server/compute/qemu/utils/qcow2.py` actually parses the qcow2 header directly.  Would you prefer to get the virtual_size and encrypted status from this, rather than shelling out to `qemu-img info` ?

grossmj:
> I noticed that gns3server/compute/qemu/utils/qcow2.py actually parses the qcow2 header directly. Would you prefer to get the virtual_size and encrypted status from this, rather than shelling out to qemu-img info ?

Sorry for the late reply. Indeed, parsing the qcow2 header would be better I think than parsing the qemu-img info :+1:

candlerb:
Great, I'll rework this in next few days and push a new version over this branch.

grossmj:
Thanks a lot and again sorry I haven't been much present, holidays time ;)

candlerb:
Patch updated.  It still shells out to `qemu-img info --backing-chain` (instead of `qemu-img rebase`) to validate the image.  But checking whether the image is encrypted, and obtaining the virtual image size, is now done directly from the header.

candlerb:
Don't apply this yet - it fails when the backing file is not qcow2 format (e.g. it's a plain image)

candlerb:
Fixed

grossmj:
I noticed the changes are for the 2.2 branch. Major changes like adding new features should be done on 3.0 however I think this is fine in this case since it is an optional feature. What do you think?

I will review soon and merge. Thanks again for this great contribution 👍

PR #1935, opened 2021-08-01
Update `remote-install.sh` with flag for KMS support and fix hostname inconsistencies

rarylson:
PR add improvements to script `remote-install.sh`:

- Add option `--without-kvm` that disables KVM support. Useful on systems that do not support nested virtualization (some cloud providers and hypervisors)
- Update GNS3 config to use newer parameters (`enable_hardware_acceleration` and `require_hardware_acceleration`)
- Fix hostname inconsistencies during script execution
- Other minor refactors

### DIFF
--- a/scripts/remote-install.sh
+++ b/scripts/remote-install.sh
@@ -23,9 +23,10 @@
 
 function help {
   echo "Usage:" >&2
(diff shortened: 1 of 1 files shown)


grossmj:
Thanks a lot, great changes 👍

PR #1970, opened 2021-09-09
Fix wrong comment for default admin password in config file

hb9cwp:


grossmj:
The removed part is for the password. Just the comment should be changed.

PR #1972, opened 2021-09-14
Add Security Onion 2.3.70, use 100GB empty .qcow2 image.

ranguli:
The version of Security Onion used in the GNS3 template (16.04) reached end of life in April. This PR adds support for the most recent version of Security Onion (2.3.70). It also switches the template over from a 30GB disk to a 100GB disk, the minimum free space necessary for an installation of Security Onion without warnings.

### DIFF
--- a/gns3server/appliances/security-onion.gns3a
+++ b/gns3server/appliances/security-onion.gns3a
@@ -24,6 +24,14 @@
     },
     "images": [
         {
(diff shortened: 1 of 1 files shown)


grossmj:
@ranguli thanks a lot for these changes however PRs for appliance files should be done on this repository: https://github.com/GNS3/gns3-registry/tree/master/appliances

The appliance files are then sync with gns3-server before each release.

Thanks 👍

PR #1998, opened 2021-11-06
Issue 1996 example plain service file

### LINKED ISSUE #1996: Example systemd service file uses legacy PID location /var/run
When starting gns3 server on our cluster, we noticed the following:
```
Nov 06 09:25:30 hel1-n15 systemd[1]: /etc/systemd/system/gns3.service:17: PIDFile= references a path below legacy directory /var/run/, updating /var/run/gns3/gns3.pid → /run/gns3/gns3.pid; please update the unit file accordingly.
Nov 06 09:25:33 hel1-n15 systemd[1]: Starting GNS3 server...
```
This referes to /var/run being deprecated in favour of /run, documented here: https://wiki.debian.org/ReleaseGoals/RunDirectory

And it is also documented more generally for the FHS:  https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html
> In FHS 3.0, /var/run is replaced by /run; a system should either continue to provide a /var/run directory or provide a symbolic link from /var/run to /run for backwards compatibility.

While /var/run is provided for compatibility, systemd still overwrites the value, so we should update it here.

I'll make a pull request to update this, but I'd like to also ask more generally why the decision was made to suggest a forking service type by default? 

Systemd can manage the process, as well as the logging to journald, so we could use a much simpler service file here as a suggested default?
I'll submit a 2nd pull request to show an example.

jak-sdk:
fixes #1996

### DIFF
--- a/init/gns3.service.systemd-plain
+++ b/init/gns3.service.systemd-plain
@@ -1,0 +1,12 @@
+[Unit]
+Description=GNS3 server
+Wants=network-online.target
+After=network.target network-online.target
+
+[Service]
+User=gns3
+Group=gns3
+ExecStart=/usr/bin/gns3server
+
+[Install]
+WantedBy=multi-user.target


grossmj:
I have added the simplified config: https://github.com/GNS3/gns3-server/commit/0d25727ca624013fd6303df82cba44ca6ecaecc0

Thanks 👍

PR #2024, opened 2022-01-17
Update qemu_vm.py linked node error

spikefishjohn:
Update the error message thrown when attempting to create a 2nd node when the template has "use as linked base VM" disabled (unchecked).


grossmj:
Thanks 👍

PR #2025, opened 2022-01-19
Drop Windows support

grossmj:


a60814billy:
I'm curious about why removing windows support in v3.0. Can you tell me the reason?

josephmhiggins:
From an amateur's perspective, a virtualization provider has to satisfy all the requirements-rules of virtualization. The only one I know of is the Popek-Goldberg Theorem. Any changes by Windows, or Apple, of hardware, i.e. CPU type, are then the responsibility of the virtualization platform, i.e. VMware, Oracle, Hyper-V, to adapt. About 5 years ago, Bernhard Ehlers brought up dropping support for VPCs and dynamips; I did not know what he was talking about then. Now, I know

grossmj:
> I'm curious about why removing windows support in v3.0. Can you tell me the reason?

To be clear, we will still support Windows but only for clients (current desktop client and web interfaces running in Chrome or Firefox). You will have to run the GNS3 server in a VM running in either VirtualBox, VMware or Hyper-V if you want to have it locally or run it on a remote Linux server / cloud instance.

We are dropping Windows support for the server for many reasons, here are a few of them:

* There are lots of issues with Windows, with firewalls and other security software getting in the way.
* No hardware acceleration for Qemu nodes (there is haxm but many appliances just cannot run efficiently with it)
* Better Docker support on Linux (to be honest I don't know how well Docker is supported on Windows at the moment)
* Dynamips doesn't run as efficiently as on Linux and it uses Cygwin which leads to some bugs.
* There are so many Linux tech we can leverage to create super interesting features like advance traffic control and Linux bridging.
* We are a small team, I believe focusing on supporting only one platform (Linux) will free us time to properly do it and give us a lot less headaches too which will hopefully translate in a better user experience.

I may have forgotten a few more...

PR #2042, opened 2022-03-23
Fix typo

kaivogelgesang:
There was a typo in `gns3server/controller/topology.py` that caused a `KeyError` when importing some appliances.


grossmj:
Thanks 👍

PR #2066, opened 2022-05-16
Add OpenRC init script

wowpetr:
This is an OpenRC init script (service) for gns3-server that supports basic commands: start, stop, restart and reload.


grossmj:
Thanks 👍

PR #2077, opened 2022-06-21
Update project.py

spikefishjohn:
Fall back to self.controller.get_compute(compute_id) if template is set to null. 

This makes it so a template housed on the main server can be used to deploy a node on a remote server using the template from the main server. The UI should prompt for which server to deploy to.

This does not address the fact that its not currently possible to set compute_id to null via the webui or that the hard drive images need to be mirrored on the remote server. 

I have done limited testing using ethernet switch. I'll do more testing tomorrow when I'm less Captain OCD on this.


spikefishjohn:
I think I got a little github over my head. I didn't mean to submit template_handler.py update on this pull request and now I can't figure out how to remove that part.

uh.. uhhh.. i'm going to go to bed and see if this fixes itself in the morning?

spikefishjohn:
I've been running with this for a few days and have seen some interesting stability issues. I'm not sure if its related or not. I'm going to pull my pull request.

PR #2083, opened 2022-07-06
fix: binding to wrong interface in windows

a60814billy:
My computer has 3 Ethernet adapter as below:
1. Ethernet (Intel Gigabit Network)
2. VMware Network Adapter VMnet1 (VMware Virtual Ethernet Adapter for VMnet1)
3. VMware Network Adapter VMnet8 (VMware Virtual Ethernet Adapter for VMnet8)

When I'm binding the interface to "Ethernet", the connection always binding to VMnet8.
I dive into the source code, and found the interface search algorithm uses "in" operator and not early return first result.
So It's runs to `"Ethernet" in interface["netcard"]` and match the last network adapter. (Because VMware Virtual **Ethernet** Adapter for VMnet8)

This PR fixes the issue, and using "exact match" in  interface["name"], then using in operator to find in interface["netcard"] if first case not found anything.

### DIFF
--- a/gns3server/compute/base_node.py
+++ b/gns3server/compute/base_node.py
@@ -778,18 +778,8 @@ class BaseNode:
             # on Linux we use RAW sockets by default excepting if host traffic must be blocked
             await self._ubridge_send('bridge add_nio_linux_raw {name} "{interface}"'.format(name=bridge_name, interface=ethernet_interface))
         elif sys.platform.startswith("win"):
-            # on Windows we use Winpcap/Npcap
-            windows_interfaces = interfaces()
-            npf_id = None
-            source_mac = None
-            for interface in windows_interfaces:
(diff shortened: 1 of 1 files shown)


josephmhiggins:
"The algorithm" has problems. The algorithm that gave me problems that I ran into is different than yours - I think. Long story short, my algorithm was after uninstalling vmware ws and switching to virtualbox. It took me 3 weeks to figure out that during the uninstalling vmware that I checked "Keep configuration settings", so I have to investigate that aspect of my problem but your code gave me a 16 hour head start on where to begin looking.

a60814billy:
Hi @grossmj, could you review this PR?
Let me know if there is anything needs to be changed.

grossmj:
> Hi @grossmj, could you review this PR?
Let me know if there is anything needs to be changed.

Looks good 👍  Sorry for the delay

PR #2086, opened 2022-07-16
Support user defined loader/libraries to run IOU

grossmj:


gns3server/compute/iou/iou_vm.py
```diff
@@ -219,6 +225,28 @@ def _check_requirements(self):
         if not os.access(self._path, os.X_OK):
             raise IOUError(f"IOU image '{self._path}' is not executable")
 
+        # set loader command
+        if elf_header_start[4] == b"\x01":
```
b-ehlers, line 229: I don't think that works, at least not with python 3.9

An array element of a byte string is an int and comparing that with a byte string (even when that contains only one element) is false.

```
>>> b"\x01\x02\x03"[1]
2
>>> type(b"\x01\x02\x03"[1])
<class 'int'>
>>> b"\x01\x02\x03"[1] == b'\x02'
False
```

So either use `if elf_header_start[4] == 1` or `if elf_header_start[4:5] == b"\x01"`

grossmj:
@b-ehlers I think we are ready to merge this excepting if you see anything else to change.

b-ehlers:
> @b-ehlers I think we are ready to merge this excepting if you see anything else to change.

I'm fine with the current state, lets merge it.

PR #2090, opened 2022-07-18
Update compute.py

spikefishjohn:
Remove check for open project. Without this check a remote can be rebooted and will be usable once the main server polls it again. Without this the main server would need to open a project that already uses the remote server or restart the main server's gns3 process.

### DIFF
--- a/gns3server/controller/compute.py
+++ b/gns3server/controller/compute.py
@@ -367,9 +367,6 @@ class Compute:
                 log.warning("Cannot connect to compute '{}': {}".format(self._id, e))
                 # Try to reconnect after 5 seconds if server unavailable only if not during tests (otherwise we create a ressource usage bomb)
                 if not hasattr(sys, "_called_from_test") or not sys._called_from_test:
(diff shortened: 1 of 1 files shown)


grossmj:
I am a bit worried about this, I remember we can easily end up in a loop trying to reconnect. Have you tested what happen if the compute isn't rebooted?

spikefishjohn:
I’ve been running all my push requests in my production setup and haven’t seen a problem. 

That being said I don’t think I understand what the trigger your talking about it. 

Is it the master coming up before the remote or the remotes coming up before the master? I do see the master polling servers that are down continuously, but I would assume that is the idea. I don’t see any cpu impact. 

If you can give me some pointers on what to look for I’m more then happy to kick the tires on it.

grossmj:
Ok, maybe I am too worried here, let's see how it goes ;)

https://github.com/GNS3/gns3-server/pull/2089: the 2.2 branch is going to be merged in master.

PR #2133, opened 2022-11-07
Migrate to importlib_resources

grossmj:


requirements.txt
```diff
@@ -11,5 +11,6 @@ psutil>=5.9.2,<5.10
 async-timeout>=4.0.2,<4.1
 distro>=1.7.0
 py-cpuinfo>=8.0.0,<8.1
+importlib-resources>=1.3
```
classabbyamp, line 14: this creates an unnecessary dependency for python 3.9+, because this version of `importlib-resources` has already been merged into the stdlib. With a couple small changes, compatibility for the py3.9+ stdlib `importlib.resources` can be added:

requirements.txt:
```diff
- importlib-resources>=1.3
+ importlib-resources>=1.3; python_version < '3.9'
```
all the places this module is used:
```diff
- import importlib_resources
+ try:
+     from importlib import resources as importlib_resources
+ except ImportError:
+     import importlib_resources
```

grossmj: Indeed, we will fix this.

grossmj: We use `https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files` which was only added in the stdlib in Python 3.9, so by importing `from importlib import resources as importlib_resources` first and then the back-ported importlib_resources, don't we risk to call a function that doesn't exist when using Python < 3.9? In other words, shouldn't try to import importlib-resources first and then revert to the stdlib if unsuccessful?

classabbyamp: yes, I think that would work better.

PR #2162, opened 2023-01-13
Add controller table in database

grossmj:
This is to save controller specific information like the current version, appliances ETAG and more in the future


grossmj:
I have decided to use a JSON file for these instead.

PR #2185, opened 2023-02-13
Update remote-install.sh and add welcome.py

Xatrekak:
### This pull request has three goals:

1. Update the remote-install.sh script to support Ubuntu Server LTS 22.10
2. Add the option to setup the welcome.py from GNS3-VM dialog to remote-install.sh using the --with-welcome option
3. Update welcome.py to run in on  Ubuntu Server LTS 22.10 outside of GNS3-VM

welcome.py is essentially Identical to the one included with GNS3VM with these changes:

1. refactored to wrap all the functions inside a class
2. updated the method get_ip() to work on any enviroment. It pulls the first system IP that is not a vbridge or loopback
3. New method repair_remote_install() that is callable from remote-install.sh to update the IP in gns3_server.conf and restart the gns3 service ensure Dialog always displays the same IP we wrote to gns3_server.conf.

### This is a stacked pulled request. 
The final merge assumes that welcome.py has already been pulled and and remote-install.sh pulls it in from the GNS3 master raw.githubusercontent

See [the first merge](https://github.com/GNS3/gns3-server/tree/229168367b27c040c8edfff77f3746d7aa566dbd) for a working example where remote-install.sh downloads welcome.py from my personal repo.

### DIFF
--- a/scripts/remote-install.sh
+++ b/scripts/remote-install.sh
@@ -26,6 +26,7 @@ function help {
   echo "--with-openvpn: Install OpenVPN" >&2
   echo "--with-iou: Install IOU" >&2
   echo "--with-i386-repository: Add the i386 repositories required by IOU if they are not already available on the system. Warning: this will replace your source.list in order to use the official Ubuntu mirror" >&2
+  echo "--with-welcome: Install GNS3-VM welcome.py script" >&2
   echo "--without-kvm: Disable KVM, required if system do not support it (limitation in some hypervisors and cloud providers). Warning: only disable KVM if strictly necessary as this will degrade performance" >&2
   echo "--unstable: Use the GNS3 unstable repository"
   echo "--help: This help" >&2
@@ -49,7 +50,7 @@ I386_REPO=0
 DISABLE_KVM=0
 UNSTABLE=0
 
-TEMP=`getopt -o h --long with-openvpn,with-iou,with-i386-repository,without-kvm,unstable,help -n 'gns3-remote-install.sh' -- "$@"`
+TEMP=`getopt -o h --long with-openvpn,with-iou,with-i386-repository,with-welcome,without-kvm,unstable,help -n 'gns3-remote-install.sh' -- "$@"`
 if [ $? != 0 ]
 then
   help
@@ -70,6 +71,10 @@ while true ; do
           ;;
         --with-i386-repository)
           I386_REPO=1
+          shift
+          ;;
+        --with-welcome)
+          WELCOME_SETUP=1
           shift
           ;;
         --without-kvm)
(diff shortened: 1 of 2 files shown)


grossmj:
@Xatrekak thanks a lot for your contribution :+1: I am going to review it in the next few days.

grossmj:
> Update the remote-install.sh script to support Ubuntu Server LTS 22.10

I suppose you meant Ubuntu 22.04 LTS? Also, what changes have you made to support it? Thanks :)

Xatrekak:
@grossmj I did in fact mean 22.04.
The change to welcome.py to support 22.04 was the get_ip() function.
In the latest LTS of ubuntu server the output of `ifconfig` has changed.
```
gns3@gns3:~$ ifconfig ens3 | grep "inet"
        inet 192.168.1.152  netmask 255.255.255.0  broadcast 192.168.1.255
        inet6 fe80::2a0:98ff:fe29:f9c3  prefixlen 64  scopeid 0x20<link>
```
The old command: `ifconfig eth0 | grep "inet\ addr" | cut -d: -f2 | cut -d" " -f1` no longer works because `addr` is not part of the output and the IP is output with spaces instead of `inet addr:192.168.1.152`
I switched the command to output parsable JSON so it should function regardless for output formatting. 

For remote-install.sh I had a few problems getting the original file to correctly setup the gns3 user and start the gns3server daemon on start. The `--with-welcome` option now also do the following:
```
mkdir /etc/systemd/system/getty@tty1.service.d
cat <<EOFI > /etc/systemd/system/getty@tty1.service.d/override.conf
[Service]
ExecStart=
ExecStart=-/sbin/agetty -a gns3 --noclear %I \$TERM
EOFI

chmod 755 /etc/systemd/system/getty@tty1.service.d/override.conf
chown root:root /etc/systemd/system/getty@tty1.service.d/override.conf

echo "gns3:gns3" | chpasswd
usermod --shell /bin/bash gns3
usermod -aG sudo gns3
```
This makes the gns3 user autologin without a password prompt. Sets a password so it can be access correctly, and also changes the default shell from sh to bash to be a littler more user friendly. 

The last thing remote-install.sh does with the --with-wlecome option is call repair_remote_install(self) from welcome.py.
This automatically updates `/etc/gns3/gns3_server.conf` with the same IP address we are displaying with the welcome dialog.
```
python3 -c 'import sys; sys.path.append("/usr/local/bin/"); import welcome; ws = welcome.Welcome_dialog(); ws.repair_remote_install()'
```
calls this code from welcome.py
```#!/usr/bin/env python3
    def repair_remote_install(self):
        """
        This method is only called by remote-install.sh during setup to ensure it is setting the same IP as shown by Dialog
        """        
        ip_addr = self.get_ip()
        subprocess.run(["sed", "-i", f"s/host = 0.0.0.0/host = {ip_addr}/", "/etc/gns3/gns3_server.conf"],capture_output=False)
        subprocess.run(["service", "gns3", "stop"],capture_output=False)
        subprocess.run(["service", "gns3", "start"],capture_output=False)
```

grossmj:
Great work, thanks a lot for your contribution:+1:

PR #2235, opened 2023-06-03
qemu : with "virtio-net-pci", fix the speed and duplex

Raizo62:
Hi

I believe that is easy to implement, so i saved you time :-)

I hope I kept the spirit of your code :-)


grossmj:
> I believe that is easy to implement, so i saved you time :-)

Thanks a lot 👍

PR #2272, opened 2023-08-12
Correctly install built-in appliances

grossmj:


grossmj:
platformdirs is only supported on Python >= 3.7

PR #2285, opened 2023-09-18
rbac fix: cannot add multiple time same privilege to a role

sylvain121:
On current version of PUT /v3/access/roles/{role_id}/privileges/{privilege_id}, database allow inserting multiple time same privilege for a role.
This generates an error when trying to DELETE the privilege.
This patch silent ignores request if privilege is already applied for the role.


grossmj:
Thanks 👍

PR #2288, opened 2023-09-20
Update README.md

Quanterm:
The release cycle of NixOs is around every half of the year.  On the unstable channel you have the latest gns3-server version.  Do not worry to use unstable because it mostly just the latest stable version. You can also do overlays to keep your other systems pkgs on the regular release version.


grossmj:
Thanks 👍

PR #2293, opened 2023-09-25
Fix: deleting resource from resource pool

sylvain121:
Fix: also delete resource from resources table.
If not we cannot add this resource to another pool


grossmj:
Thanks, I have adapted your fix in https://github.com/GNS3/gns3-server/commit/a27db6b4eb1f8134843ecdc9bf2bab5f4853694c 

There is no need to merge this PR anymore.

PR #2321, opened 2023-11-07
Fixed updating system and GNS3 via welcome.py.

Xatrekak:
Fixed the upgrade option under welcome.py adding 3 options. GNS3 only, all packages, and a full system update.

Also added user permissions to sudoers.d/gns3 so the update could be executed without requesting the user password.

Screenshot from 2023-11-06 19-06-05


grossmj:
Thanks 👍

PR #2345, opened 2024-01-31
Code review - Working POC of fixed telnet console using telnetlib3

spikefishjohn:
This isn't ready for PR merge. I'm just doing this more or less for a code review. 

This only changes base_node.py to use sftelnetproxymuxer. 

Adds a different queue for remote clients (set).

Deletes clients from set when they're bad.

Sends heartbeats using telnet NOP packets to verify remote clients are still up. 

Forgot to check if QEMU is ok with this on the server side. At one point it was making garbage on the remote clients sessions, but I might have been doing something wrong.

Can recover from a dead telnet session, so hurray for that.

Asyncio is hard.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-20.04  # Downgrade Ubuntu to 20.04 to fix missing Python 3.6
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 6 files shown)


spikefishjohn:
oh .. telnetlib3 requires python 3.7 or higher.. so .. uh.. I removed python 3.6 in this PR to get the unit tests to fire off.

spikefishjohn:
I just noticed this includes changes to telnet_server.py. This shouldn't be included. I'm trying to see if I can edit the PR to remove that, otherwise i'll cancel.

spikefishjohn:
Closing this. Seems like it can't easily be done through webui.

PR #2346, opened 2024-01-31
New telnet proxy console poc

spikefishjohn:
This isn't ready for PR merge. I'm just doing this more or less for a code review.

This only changes base_node.py to use sftelnetproxymuxer.

Adds a different queue for remote clients (set).

Deletes clients from set when they're bad.

Sends heartbeats using telnet NOP packets to verify remote clients are still up.

Forgot to check if QEMU is ok with this on the server side. At one point it was making garbage on the remote clients sessions, but I might have been doing something wrong.

Can recover from a dead telnet session, so hurray for that.

Asyncio is hard.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-20.04  # Downgrade Ubuntu to 20.04 to fix missing Python 3.6
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


spikefishjohn:
If you wanted to go this way i just noticed something. 

https://github.com/jquast/telnetlib3/blob/5fbc6f723a6cb37651e782f6b35bd7e6cebabd5a/telnetlib3/client_shell.py#L19

The telnet client built into telnetlib3 doesn't support windows. I think that would make using telnetlib3 a 3.0 only thing.

Issue #2344, opened 2024-01-31
Telnet console death issue


spikefishjohn:
@grossmj 

Since there isn't really an open issue to track the original death of the telnet console I'm opening this. I've narrowed down where the issue is. I haven't figured out what to do about it just yet.

Here is a link to the telnet_server.py with debugs showing the issue.
[telnet_server.py](https://github.com/SpikefishSolutions/gns3-server/blob/b759d31482aeb28033c2d25af3da256e08a55a0d/gns3server/utils/asyncio/telnet_server.py)

Here is the debug logs when the death loop happens.

```
2024-01-30 23:13:41 DEBUG telnet_server.py:271 __process True loop
2024-01-30 23:13:41 DEBUG telnet_server.py:273 __process  reader_read is None
2024-01-30 23:13:41 DEBUG telnet_server.py:251 Start async _get_reader
2024-01-30 23:13:41 DEBUG telnet_server.py:256 self._reader_process == network_reader
2024-01-30 23:13:41 DEBUG telnet_server.py:257 <StreamReader waiter=<Future pending cb=[Task.task_wakeup()] created at /usr/lib/python3.10/asyncio/base_events.py:429> transport=<_SelectorSocketTransport fd=11 re
ad=polling write=<idle, bufsize=0>>> == <StreamReader waiter=<Future pending cb=[Task.task_wakeup()] created at /usr/lib/python3.10/asyncio/base_events.py:429> transport=<_SelectorSocketTransport fd=13 read=poll
ing write=<polling, bufsize=65550>>>
2024-01-30 23:13:41 DEBUG telnet_server.py:262 _get_reader Returning None
2024-01-30 23:13:41 DEBUG telnet_server.py:275 __process reader_read is still 2nd None
2024-01-30 23:13:41 DEBUG telnet_server.py:277 __process reader_read is still 3rd None
2024-01-30 23:13:42 DEBUG telnet_server.py:292 __process just before coro done check
2024-01-30 23:13:42 DEBUG telnet_server.py:293
2024-01-30 23:13:42 DEBUG telnet_server.py:294
2024-01-30 23:13:42 DEBUG telnet_server.py:295
2024-01-30 23:13:42 DEBUG telnet_server.py:271 __process True loop
2024-01-30 23:13:42 DEBUG telnet_server.py:273 __process  reader_read is None
2024-01-30 23:13:42 DEBUG telnet_server.py:251 Start async _get_reader
[...]

PR #2347, opened 2024-02-03
Address the telnet console death bug. 

spikefishjohn:
Add wait_for for drain() call with a 5 second timeout. If we're stuck on drain then the buffer isn't getting emptied. 5 seconds after drain() blocks, exception will be thrown and client will be removed from connection table and will no longer be a problem. There will be a pause in console output while in this 5 second wait, after that the console will return to normal. 

When the exception is triggered the follow debug will be logged.

2024-02-03 03:07:15 DEBUG telnet_server.py:309 Timeout while sending data to client: ('REMOTE_IP', REMOTE_PORT), closing and removing from connection table. 

It just dawned on me that peerinfo() is a list so if you want that reformatted it would be easy enough to do. 

Fixes [Issue 2344](https://github.com/GNS3/gns3-server/issues/2344)

### DIFF
--- a/gns3server/utils/asyncio/telnet_server.py
+++ b/gns3server/utils/asyncio/telnet_server.py
@@ -297,9 +297,17 @@ class AsyncioTelnetServer:
                     reader_read = await self._get_reader(network_reader)
 
                     # Replicate the output on all clients
-                    for connection in self._connections.values():
-                        connection.writer.write(data)
-                        await connection.writer.drain()
+                    for connection_key in list(self._connections.keys()):
+                        client_info = connection_key.get_extra_info("socket").getpeername()
+                        connection = self._connections[connection_key]
+
+                        try:
+                            connection.writer.write(data)
+                            await asyncio.wait_for(connection.writer.drain(), timeout=10)
+                        except:
+                            log.debug(f"Timeout while sending data to client: {client_info}, closing and removing from connection table.")
+                            connection.close()
+                            del self._connections[connection_key]
 
     async def _read(self, cmd, buffer, location, reader):
         """ Reads next op from the buffer or reader"""


spikefishjohn:
oops, just noticed its a 10 second timeout and not 5.

cristian-ciobanu:
@spikefishjohn so merging this pull request will fix completely the console freeze bug when using telnet ?

spikefishjohn:
It does for me. May want to lower the timeout as 10 seconds feels a little long. That being said the patch could use additional testing. Do you have a way to replicate the issue btw?

The way I recreate the issue is clear the firewall's connection table for the telnet session (I'm accessing GNS3 through a firewall to be clear). This will cause the firewall to do a silent drop on the tcp session. Then open a 2nd telnet session and do something that creates a large mount of output on the console. I use find / on a linux node for example. 

Without the patch the console will output data for about 10 seconds or so then lock up. You'll be able to open new sessions and get connected and even send data but nothing will come back.

With the patch there will be a pause while GNS3 is trying to communicate with the down telnet session. After 10 seconds it will give up, delete the session and then session 2 will start working again as well as any new session.

Also note the next version of GNS3 will have tcp keepalives timers lowered so that it sends keep alives more often then every 2 hours. This will prevent a firewall from killing a session from being idle for too long but won't fix the broken console issue.

cristian-ciobanu:
> Do you have a way to replicate the issue btw?

I can only replicate this only by leaving the telnet console session opened for a while (no activity for several hours I think) until I receive the message connection refused.

spikefishjohn:
hmmmm.. connection refused. Ok I can't say for sure this will address your issue but only because I don't know what happens if the telnet console is left in a bad state for long enough. I personally have never seen it throw connection refused, which would mean the listening socket closed. That being said if the GNS3 node continues to try to push data out and it can't write data at some point I would expect the service to do something bad (like crash) ... so its possible.

If you can give me more details i'll see if I can replicate or by all means give the update a try and report back.

For me to replicate can you tell me how your accessing the telnet console network topology wise? To give you an idea of bullet points i'm looking for.

* GSN3 version
* client OS
* GNS3 server OS
* GNS3 node your connecting to (cisco router, linux, etc). Hopefully its something I have access to.
* How does the client reach the server? (its accessed through router, firewall, VPN, GNS3 VM which is running on Virtbox/Vmware/ESX etc (crossing fingers its not GNS3VM).
Just be as detailed as possible if you want me to try to replicate. Minor details may not be. :)

cristian-ciobanu:
Sorry I got a bit confused. I have two different issues related to telnet console connections.

1. The telnet console session freezes if I leave opened for a while (no activity for several hours I think) no message received. 
2. When I use auto-start console option for devices the console of the devices automatically pops in with the message connection refused (using SecureCRT for accessing console of devices). If I close the console and reopen again it starts to work fine.

I have two different setups for GNS3.
1. GNS3 server 2.2.44 running on bare metal Ubuntu 22.04 machine and GNS3 client 2.2.44 running remotely on Windows.
2. GNS3 3.0.0 beta (server and client) running on bare metal on Linux (same machine Archlinux)

The first issue with the telnet console freeze happens only with the remote server setup not when running both server and client on same machine.
The second issues happens randomly so it is not easy to replicate. I think maybe when I start multiple devices at once and their all open their console, but I need to do some more extensive testing.

spikefishjohn:
ah ok.

For sure, this will fix issue number 1. I'll be my burrito on it. Please give the patch a try to see if I owe you one.

Issue 2 sounds like a completely different thing. Like a timing issue where it hasn't waited for the device to return running or maybe for the console to have started. I would make a new bug report for that.

cristian-ciobanu:
@grossmj 3.0 branch does not need this patch applied or the other pull request using telnetlib3 rewrite should in theory fix this annoying issue ?

grossmj:
> @grossmj 3.0 branch does not need this patch applied or the other pull request using telnetlib3 rewrite should in theory fix this annoying issue ?

The patch will be merged into 3.0 and the other pull request using telnetlib3 should make it there as well once it has been sufficiently tested.

spikefishjohn:
sooo.. uh.. loaded up 2.2.46. Opened a telnet console. Slept laptop, went to bed. Work up. Double click vyos box.

... nothing...

.....

image

spikefishjohn:
I did notice something. [this](https://github.com/GNS3/gns3-server/blob/5bab4131e186fbaa08943987fa9f7deb6a409642/gns3server/utils/asyncio/telnet_server.py#L306) calls except if it fails which calls [close()](https://github.com/GNS3/gns3-server/blob/5bab4131e186fbaa08943987fa9f7deb6a409642/gns3server/utils/asyncio/telnet_server.py#L309) which then does this.
```python
   async def close(self):
        for writer, connection in self._connections.items():
            try:
                writer.write_eof()
                await writer.drain()
                writer.close()
                # await writer.wait_closed()  # this doesn't work in Python 3.6
            except (AttributeError, ConnectionError):
                continue
```

So my theory is the drain call needs to have a wait_for as well. I'm going to see if I can recreate the problem then take a swing at it again.

PR #2413, opened 2024-08-23
Duplicate faster

z5241248:
Copying project files directly, rather than copying them in an import-export fashion, can make copying projects many times faster


gns3server/controller/project.py
```diff
@@ -1237,3 +1247,66 @@ def __json__(self):
...
+        # work dir
+        p_work = pathlib.Path(location or self.path).parent.absolute()
+        t0 = time.time()
+        new_project_id = str(uuid.uuid4())
+        new_project_path = p_work.joinpath(new_project_id)
+        # copy dir
+        scripts_path = os.path.join(pathlib.Path(__file__).resolve().parent.parent.parent, 'scripts')
+        process = await asyncio.create_subprocess_exec('python', os.path.join(scripts_path, 'copy_tree.py'), '--src',
+                                                       self.path, '--dst',
+                                                       new_project_path.as_posix())
```
grossmj, line 1268: Why do you call an external Python script to copy the tree and not shutil.copytree() directly?

z5241248: Indeed, we can use shutil.copytree(), but it will cause the entire program request to be blocked. For example, the page cannot be refreshed in time. So, I think the external Python script may be a feasible plan.

If you have a better plan, that would be better.

z5241248: I thought about it, it's my business requirements
. If you don't mind requesting blocking, I can change the code, using shutil.copytree() is also no problem.

z5241248: Thank you, I'll try it. If it works, I'll submit it again.

z5241248: It works. I pushed it again. Thanks : )

gns3server/controller/project.py
```diff
@@ -1237,3 +1251,62 @@ def __json__(self):
...
+        topology["auto_start"] = False
+        topology["auto_open"] = False
+        topology["auto_close"] = False
+        # change node ID
+        node_old_to_new = {}
+        for node in topology["topology"]["nodes"]:
+            new_node_id = str(uuid.uuid4())
+            if "node_id" in node:
+                node_old_to_new[node["node_id"]] = new_node_id
+                _move_node_file(new_project_path, node["node_id"], new_node_id)
```
grossmj, line 1286: There is an Unresolved reference '_move_node_file' for this line.

z5241248: Sorry, I think someone must change the code, I'll check and re-push it tomorrow morning.

grossmj: Thanks, I have fixed it in https://github.com/GNS3/gns3-server/commit/19bd953d31a9c80414fa53baec54181e934ffefd

grossmj: I am testing it now 👍

grossmj:
I have made some minor changes. I will do some more tests tomorrow and I expect to merge your code after that. Thanks 👍

z5241248:
> I expect to merge your code

In addition to the code, I noticed you have corrected a few grammatical errors.

My native language is not English, so thank you for correcting the code. That's great.

Expect good results.

z5241248:
> I have made some minor changes. I will do some more tests tomorrow and I expect to merge your code after that. Thanks 👍

What are the test results now?

grossmj:
Merged! thanks again for this contribution! 👍

z5241248:
> Merged! thanks again for this contribution! 👍

Hello, I asked you a question on the GNS3 web message, could you reply to me when you have time?

PR #2471, opened 2024-12-31
Watchdog migration

grossmj:


grossmj:
The migration from watchfiles to watchdog was necessary to detect the 'closed' event which is not supported by watchfiles. The closed event ensures we don't try to add an image to the controller database while the image has finished to be copied or uploaded to the images directory.

PR #2523, opened 2025-04-13
Update Dockerfile to use Python 3.10 instead of Python 3.8

SystemVll:
### Issue
The GNS3 server package requires Python >=3.9, but our Dockerfile was using the default Python 3.8.10 from Ubuntu Focal. This was causing build failures with the error:
```
ERROR: Package 'gns3-server-3.0.5.dev1' requires a different Python: 3.8.10 not in '>=3.9'
```

### Changes
- Removed the installation of system Python packages (`python3-pip`, `python3-all`, etc.)
- Added build steps to compile and install Python 3.10.16 from source
- Set up pip for Python 3.10
- Updated all Python commands to explicitly use Python 3.10

### Benefits
- Resolves the compatibility issue with GNS3 server requirements
- Provides a consistent Python environment across deployments

These changes allow the Docker image to build successfully and run the GNS3 server with the required Python version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,14 +15,21 @@ COPY ./requirements.txt /gns3server/requirements.txt
 RUN DEBIAN_FRONTEND=noninteractive apt install -y \
     locales \
     software-properties-common \
-    python3-pip \
-    python3-all \
-    python3-setuptools \
-    python3-dev \
     busybox-static \
     gcc \
     qemu-kvm \
     libvirt-daemon-system
+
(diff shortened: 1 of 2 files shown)


grossmj:
Thanks for this. I have just one question though, why just not upgrade to use ubuntu:noble instead?

Inplex-sys:
I choose to upgrade Python only to prevent any possible disruption to the overall environment. I didn't upgrade the base image to use Ubuntu Noble because it might involve possible compatibility issues with other dependencies or configurations.

grossmj:
Thanks for your contributing :+1: however I have added "ubuntu:noble" in the Dockerfile. I think it is a better solution. Even though I haven't merged your PR, please do not hesitate to open more in the future. Thanks!

Issue #2568, opened 2025-10-17
Database error when using multiple clients on a single GNS3 server


gigibu5:
Hello,
for our Secondary school's Network class LAB exercises we are using GNS3 as our network simulator tool. The transission from Cisco packet tracer was very much appreciated. For the whole classroom we are using a single GNS3 server instance, beacuse some LAB exercises require group work using Cloud and a local Bridge.

Until this week we were using GNS3 v2, which worked great but decided to update to v3 beacuse of the RBAC. 

After updating to version 3.0.5 we have experienced a live demo problem. The GNS3 clients were returning timeout errors and a messeage to view the logs for more infomration. The logs were filled with the following error:
```
2025-10-15 10:02:59 ERROR server.py:199 Controller database error in /v3/projects (POST): QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/20/3o7r)
```
The problem is with the db connection client library, which has an upper limit of 5 simultanious connections to the database (not even close to the required amount of connections for a classroom of 16 students). 

**Hotfix**: As an in lab quick fix I searched through the file system and found the engine_create function, which has a default value of `pool_size` set to 5 from the QueuePool class constructor. I have edited the folowing 2 lines in the class constructor (file: `/usr/share/gns3/gns3-server/lib/python3.12/site-packages/sqlalchemy/pool/impl.py`)
```py
def __init__(
        self,
        creator: Union[_CreatorFnType, _CreatorWRecFnType],
        pool_size: int = 500, # old value -> 5
        max_overflow: int = 1000, # old value -> 10
        timeout: float = 30.0,
        use_lifo: bool = False,
        **kw: Any,
):
```
The fix solved all the problem's, but this **is not a proper solution to the problem**.

Thank you in advance for looking into the issue.

Issue #2566, opened 2025-10-17
Exporting projects with non-ASCII names


gigibu5:
When using non-ASCII (for example č, š, ž, Č, Š and Ž from the Slovenian Alphabet) in project names and trying to export the project using the `File`-> `Export project` utility the GNS3 server responds with `Internal server error (HTTP code 500)`. This was not the problem in GNS3 v2. In the mean time the issue is fixable by using only ASCII charachters in project names.

I am attaching an image of the error pop-up:
<img width="717" height="212" alt="Image" src="https://github.com/user-attachments/assets/19731230-9293-492b-8137-c42c745f5aaf" />

Thank you in advance for looking into the issue.

PR #2569, opened 2025-10-17
Fix database pool error and non-ASCII character project export

Mytic2330:
#2568
The first patch fixes the above specified issue.
In `/db/tasks.py` where the initialization of the database is handeled, we have adjusted the code in the following way:

Changed code:
```py
engine = create_async_engine(db_url, connect_args={"check_same_thread": False, "timeout": 20}, future=True, pool_size=512, max_overflow=1024)
```

We have introduced the `pool_size` and `max_overflow` parameters to overwrite the default 5 and 10 values. 
The code was tested and allowed more than 20 clients to connect without any errors raised.

Co-Authored-By: @UfiKing

#2566
The second patch fixes the above specified issue.
In `/api/routes/controller/projects.py` where the export of the project is handeled, we have adjusted the code in the following way:
```py
fallback = project.name.encode("ascii", "ignore").decode() or "project"
encoded = urllib.parse.quote(project.name, safe="")
headers = {
    "Content-Disposition": (
        f'attachment; filename="{fallback}.gns3project"; filename*=UTF-8\'\'{encoded}.gns3project'
    )
}
```
Example of handling a non-ASCII project name:
<table>
<tr>
 <td>Original
 <td>Fallback
 <td>Encoded
<tr>
 <td>"Tešt"
 <td>"Tet"
 <td>"Te%C5%A1t"
</table>

`fallback` would remove all non-ASCII characters.
`encoded` would UTF-8 percent encode the non-ASCII characters.
> [!NOTE]
>  Both are provided so you can still read the original name on clients that support it or you get the reduced name (or the "project" if all the characters in the project name were non-ASCII) on the clients that do not.

Co-Authored-By: @UfiKing

### DIFF
--- a/gns3server/api/routes/controller/projects.py
+++ b/gns3server/api/routes/controller/projects.py
@@ -380,7 +380,13 @@ async def export_project(
     except (ValueError, OSError, RuntimeError) as e:
         raise ConnectionError(f"Cannot export project: {e}")
 
-    headers = {"CONTENT-DISPOSITION": f'attachment; filename="{project.name}.gns3project"'}
+    fallback = project.name.encode("ascii", "ignore").decode() or "project"
+    encoded = urllib.parse.quote(project.name, safe="")
+    headers = {
+        "Content-Disposition": (
+            f'attachment; filename="{fallback}.gns3project"; filename*=UTF-8\'\'{encoded}.gns3project'
+        )
+    }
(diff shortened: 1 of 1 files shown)


Mytic2330:
## Tested:
When exporting the project on web-ui the project name appears correctly in the downloads even with non-ASCII characters if the browser supports it.

image

PR #2616, opened 2026-02-21
Add 'created_by' field to track project creator

StupidTreee:
I'm building an exam environment with gns3 for my diploma thesis, which required to know who created certain projects for automated role allocation. since that information isn't stored this is the solution i came up with. I'm not sure if other people might have use cases for this but i thought i might contribute it anyway since it seems like useful information.

Please note that this is my first PR ever in a public repo so there's probably some things i missed.

When a project is created via the API, the 'created_by' field is automatically set to the authenticated user's username. This allows identifying who created each project.

Changes:
- Add 'created_by' parameter to Project model (init, property, asdict)
- Add 'created_by' to ProjectBase and Topology pydantic schemas
- Inject current user as 'created_by' in create_project API route
- Persist 'created_by' in .gns3 topology files
- Add test_created_by and update test_json

### DIFF
--- a/gns3server/api/routes/controller/projects.py
+++ b/gns3server/api/routes/controller/projects.py
@@ -123,8 +123,7 @@ async def create_project(
 
     controller = Controller.instance()
     project_dict = jsonable_encoder(project_data, exclude_unset=True)
-    if "created_by" not in project_dict:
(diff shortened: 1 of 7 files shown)


grossmj:
Thanks for your PR!

PR #2618, opened 2026-02-25
feat(virtualbox): add fallback for VBoxManage executable detection

yueguobin:
## Problem Description

When updating a node through the UI (e.g., by removing tags), the server would crash with a `TypeError: argument of type 'NoneType' is not iterable` error. The issue occurred when the client sent `custom_adapters` data with `port_name` set to `null`.

## Root Cause

The problem occurred in three locations where the code retrieves custom port names from adapter settings:

1. When using the `.get()` method with a default value, if the key exists but has a `None` value, the method returns `None` instead of the default value
2. The code then attempts to create Port objects with `name` set to `None`
3. When the Port's `short_name` property is accessed, it tries to perform string operations on `None`, causing the error

### 1. Port Class Enhancement

Updated the `short_name` property in the Port base class to handle cases where the port name is `None`. Added a check to ensure string operations are only performed when the name exists, preventing `NoneType` errors during string manipulation.

### 2. Node Class Docker Port Handling

Modified how Docker node port names are retrieved from custom adapter settings. Instead of relying solely on the default value mechanism, the code now uses a logical OR operation to ensure a valid port name is always provided. If the custom port name is `None` or empty, the default port name format is used.

### 3. Port Factory Standard Port Handling

Updated the StandardPortFactory in two locations where custom port names are applied:
- First port handling: Ensures the first port name falls back to the configured default if the custom name is `None`
- Subsequent ports: Ensures formatted port names are preserved when custom names are `None` or empty

## Impact

This fix prevents server crashes when:
- Clients send `custom_adapters` with `null` port names during node updates
- Port data loaded from databases contains missing or null port names
- Any other scenario where port names might be `None`

The fix is defensive and ensures that the server can handle malformed or incomplete port configuration data gracefully, using sensible defaults instead of crashing.

### DIFF
--- a/gns3server/compute/virtualbox/__init__.py
+++ b/gns3server/compute/virtualbox/__init__.py
@@ -68,6 +68,8 @@ class VirtualBox(BaseManager):
                     vboxmanage_path = vboxmanage_path_osx
             if not vboxmanage_path:
                 vboxmanage_path = shutil.which("vboxmanage")
+            if not vboxmanage_path:
+                vboxmanage_path = shutil.which("VBoxManage")
 
         if vboxmanage_path and not os.path.exists(vboxmanage_path):
             log.error(f"VBoxManage path '{vboxmanage_path}' doesn't exist")
--- a/gns3server/controller/node.py
+++ b/gns3server/controller/node.py
@@ -761,7 +761,8 @@ class Node:
                     if custom_adapter["adapter_number"] == adapter_number:
                         custom_adapter_settings = custom_adapter
                         break
-                port_name = custom_adapter_settings.get("port_name") or f"eth{adapter_number}"
+                port_name = f"eth{adapter_number}"
+                port_name = custom_adapter_settings.get("port_name", port_name)
                 mac_address = custom_adapter_settings.get("mac_address")
                 if not mac_address and "mac_address" in self._properties:
                     mac_address = int_to_macaddress(macaddress_to_int(self._properties["mac_address"]) + adapter_number)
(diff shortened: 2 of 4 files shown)


yueguobin:
Submitted in conjunction with https://github.com/GNS3/gns3-web-ui/pull/1579

yueguobin:
```
Request URL http://127.0.0.1:3080/v3/projects/27460c98-3aaf-465d-aaef-c62c0bd6a085/nodes/fb88b952-2f21-497d-99de-f5936540984d
Request Method PUT
```

image

I'm not sure if port_name: null is an issue with the Web UI implementation. Does the desktop UI send data in this format when updating node information?

grossmj:
> I'm not sure if port_name: null is an issue with the Web UI implementation. Does the desktop UI send data in this format when updating node information?

The desktop Ui generate a list of custom adapters with default port names already set. It never sends empty `port_name`.

Screenshot from 2026-02-26 21-43-59

Changing the name and type of the last port sends the following data to update the node:

```
{'compute_id': 'local', 'node_type': 'qemu', 'node_id': 'c6d2ff78-4745-4d74-b071-7e371ab59475', 'properties': {'custom_adapters': [{'adapter_number': 15, 'port_name': 'test', 'adapter_type': 'i82551'}]}}

```

yueguobin:
The fix is limited to https://github.com/GNS3/gns3-server/pull/2618/commits/c55fd9e06606996265faf5867e4ba405a55e9243.

Issue #353, opened 2015-11-10
GNS3 Python API


noplay:
_From @grossmj on November 5, 2015 21:27_

It would be great to have a python module to access GNS3 topology control, like:
 
- shut down / bring up interfaces
- reload / stop / start devices
- build connections between device interfaces
- Control link quality, even during topology running (emulate link quality degradation)
mininet-like:

Emulated Links. The data rate of each link is enforced by Linux Traffic Control (tc), which has a number of packet schedulers to shape traffic to a configured rate...
ex: self.addLink(host, switch, bw=10, delay='5ms', loss=10, max_queue_size=1000)

- add devices to the topology

_Copied from original issue: GNS3/gns3-gui#790_

noplay:
It's partially done with the REST Api. But the management of multiple servers make it hard to use for users.

boenrobot:
I wouldn't mind a similar feature for importing/exporting configs out of VMs.

Because each VM could potentially be any type of guest OS, it would be great if the user could instruct GNS3 on what the guest OS is, and then based on that, enable the import/export of configurations.

3rd parties should be able to register one or more new types of guest OS-es that they can import/export configs for, and have those handlers executable from the GUI.

noplay:
Yes good remark

Le dim. 27 déc. 2015 18:58, Vasil Rangelov <notifications@github.com> a
écrit :

grossmj:
We need a higher lever API which will come with https://github.com/GNS3/gns3-server/issues/417

mvdwoord:
Hi, I was looking for this functionality and decided to give it a try to build something myself. It is very much just a beginning at the moment, only implemented some basic GET requests basically. I am not by any means a seasoned developer, only been using GNS3 for a week or two but figured it would be at the very least a useful experience for myself.

If any of you could have a look at [the example notebook](https://github.com/mvdwoord/PyGNS3/blob/master/Example.ipynb) and let me know what you think.. that would be greatly appreciated. [Link to full repo](https://github.com/mvdwoord/PyGNS3)

If this is vaguely in the direction you have in mind I would be more than willing to invest some time in the upcoming weeks to build it out... otherwise I will focus just on scratching my own itch with this and stay out of your way.

Thanks!
Maarten

grossmj:
Hi @mvdwoord, we definitely welcome your contribution

I think this is a great start, keep it simple for users and clean :+1: 

Let us know if you have questions.

mvdwoord:
[...]

PR #2623, opened 2026-03-01
Feature/user settings api

yueguobin:
# User Settings API

## Overview

The User Settings API provides a unified and flexible configuration management system. User configurations are stored directly in the `users` table as a JSON field, eliminating the need for separate tables.

---

### 1. Profile-Based Configuration

Store all user settings as profiles. Each profile represents a complete configuration that can include:
- LLM provider settings
- API credentials
- Model parameters
- Any custom fields via extensible JSON storage

**Use Cases:**
- Multiple LLM configurations for different providers
- Quick switching between work and personal settings
- A/B testing different model parameters
- Custom configurations for specific projects

### 2. Active Profile Management

Users can create multiple profiles and mark one as "active". The system always uses the active profile for operations.

### 3. Extensible Field Support

Beyond core fields, the API accepts any additional custom fields.

---

### Storage Location

All data is stored in the GNS3 controller database (SQLite):

### Storage Structure

Configurations are stored as a JSON field `model_configs` in the `users` table:

```
users table:
┌────────────────────────────────────┐
│ user_id                            │
│ username                           │
│ email                              │
│ ...                                │
│ model_configs (TEXT/JSON)          │
│   {                                │
│     "profiles": [...],             │
│     "active": "profile_name"       │
│   }                                │
└────────────────────────────────────┘
```

This design:
- Eliminates the need for a separate `user_settings` table
- Simplifies queries (no joins required)
- Reduces code complexity
- Maintains flexibility for future fields

---

## API Endpoints

| Method | Endpoint | Description |
|--------|----------|-------------|
| GET | `/v3/access/users/{user_id}/profiles` | Get all profiles |
| POST | `/v3/access/users/{user_id}/profiles` | Create a new profile |
| PUT | `/v3/access/users/{user_id}/profiles/{name}` | Update a profile |
| DELETE | `/v3/access/users/{user_id}/profiles/{name}` | Delete a profile |
| GET | `/v3/access/users/{user_id}/profiles/active` | Get the active profile |
| PUT | `/v3/access/users/{user_id}/profiles/active` | Set the active profile |

---

## Authentication

All endpoints require JWT authentication:

```
Authorization: Bearer <your_token>
```

Tokens are obtained via `/access/users/login`.

---

### ⚠️ API Key Storage

**API keys are stored in PLAINTEXT** in the database. This is a security consideration that requires proper handling:

**Database Security:**
- 🔒 **Protect the database file** - Ensure proper file permissions (e.g., `chmod 600` on Linux)
- 🔒 **Limit database access** - Only the GNS3 service account should have read access
- 🔒 **Regular backups** - Backup the database securely and store in encrypted form
- 🔒 **Monitor access** - Audit and monitor database file access

**Network Security:**
- 🔒 **Use HTTPS in production** - Encrypt API keys in transit
- 🔒 **Secure internal communication** - Use TLS between services
- 🔒 **VPN/Private networks** - Deploy API endpoints in trusted networks only

**Access Control:**
- ✅ **RBAC enforced** - All endpoints require proper permissions
- ✅ **JWT authentication** - Valid token required for all operations
- ✅ **Audit logging** - Track who accesses which profiles

**Data at Risk:**
```
users.model_configs (JSON):
{
  "profiles": [
    {
      "api_key": "sk-xxxxx",  ← Stored in plaintext!
      ...
    }
  ]
}
```

**Recommendations:**
1. **Never commit database files** to version control
2. **Use environment-specific credentials** - Different keys for dev/staging/prod
3. **Rotate API keys regularly** - Implement key rotation policies
4. **Consider encryption at rest** - For highly sensitive deployments, consider database encryption
5. **Limit token lifespan** - Use short-lived JWT tokens when possible

---

### Core Fields (Required)

| Field | Type | Required | Default | Description |
|-------|------|----------|---------|-------------|
| name | string | Yes | - | Unique profile identifier (1-50 chars) |
| provider | string | No | "openai" | LLM provider name |
| model | string | Yes | - | Model name/identifier |
| api_key | string | Yes | - | API authentication key |
| base_url | string | No | "" | API endpoint URL (empty = use provider default) |
| temperature | string | No | "0.7" | Generation temperature |

### Naming Restrictions

**Reserved names:**
- ⚠️ `"active"` - Reserved for system use (conflicts with `/profiles/active` route)

Attempting to create or rename a profile to `"active"` will result in an error:
```json
{
  "detail": "Profile name 'active' is reserved for system use"
}
```

### Extended Fields (Optional)

Any additional fields are accepted and stored:

- `max_tokens` - Maximum response tokens
- `top_p` - Nucleus sampling parameter
- `stream` - Enable streaming
- Custom provider-specific fields

### Provider Default URLs

When `base_url` is empty or not provided, the client should use provider-specific defaults:

| Provider | Default Base URL |
|----------|------------------|
| openai | https://api.openai.com/v1 |
| qwen | https://dashscope.aliyuncs.com/compatible-mode/v1 |
| anthropic | https://api.anthropic.com |
| deepseek | https://api.deepseek.com/v1 |
| moonshot | https://api.moonshot.cn/v1 |
| zhipu | https://open.bigmodel.cn/api/paas/v4 |

**Note**: Clients are responsible for applying default URLs when `base_url` is empty.

---

### Create a Profile

```bash
curl -X POST "http://localhost:3080/v3/access/users/{user_id}/profiles" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "openai-default",
    "provider": "openai",
    "model": "gpt-4",
    "api_key": "sk-xxx"
  }'
```

Note: `base_url` is optional. If not provided, the client will use the provider's default URL.

### Create a Profile with Custom Base URL

```bash
curl -X POST "http://localhost:3080/v3/access/users/{user_id}/profiles" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "qwen",
    "provider": "qwen",
    "model": "qwen-max",
    "api_key": "sk-xxx",
    "base_url": "https://api.qwen.com/v1",
    "temperature": "0.7",
    "max_tokens": 2000
  }'
```

### Create a Profile with Extended Fields

The API accepts any additional custom fields. These are saved and returned as-is:

```bash
curl -X POST "http://localhost:3080/v3/access/users/{user_id}/profiles" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "deepseek",
    "provider": "deepseek",
    "model": "deepseek-chat",
    "api_key": "sk-xxx",
    "temperature": "0.0",
    "base_url": "https://api.deepseek.com/v1",
    "max_tokens": 4000,
    "top_p": "0.9",
    "custom_field": "custom_value"
  }'
```

**Response:**
```json
{
  "name": "deepseek",
  "provider": "deepseek",
  "model": "deepseek-chat",
  "api_key": "sk-xxx",
  "temperature": "0.0",
  "base_url": "https://api.deepseek.com/v1",
  "max_tokens": 4000,
  "top_p": "0.9",
  "custom_field": "custom_value"
}
```

### Get All Profiles

```bash
curl -X GET "http://localhost:3080/v3/access/users/{user_id}/profiles" \
  -H "Authorization: Bearer <token>"
```

**Response (with profiles):**
```json
{
  "profiles": [
    {
      "name": "qwen",
      "provider": "qwen",
      "model": "qwen-max",
      "api_key": "sk-xxx",
      "base_url": "https://api.qwen.com/v1",
      "temperature": "0.7",
      "max_tokens": 2000
    }
  ],
  "active": "qwen"
}
```

**Response (empty profiles):**
```json
{
  "profiles": [],
  "active": "default"
}
```

### Set Active Profile

```bash
curl -X PUT "http://localhost:3080/v3/access/users/{user_id}/profiles/active" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{"profile_name": "qwen"}'
```

### Get Active Profile

```bash
curl -X GET "http://localhost:3080/v3/access/users/{user_id}/profiles/active" \
  -H "Authorization: Bearer <token>"
```

**Response:**
```json
{
  "name": "qwen",
  "provider": "qwen",
  "model": "qwen-max",
  "api_key": "sk-xxx",
  "base_url": "https://api.qwen.com/v1",
  "temperature": "0.7",
  "max_tokens": 2000
}
```

### Update a Profile

```bash
curl -X PUT "http://localhost:3080/v3/access/users/{user_id}/profiles/qwen" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -d '{
    "temperature": "0.9",
    "max_tokens": 4000
  }'
```

### Delete a Profile

```bash
curl -X DELETE "http://localhost:3080/v3/access/users/{user_id}/profiles/qwen" \
  -H "Authorization: Bearer <token>"
```

---

## Error Handling

| Status | Description |
|--------|-------------|
| 200 OK | Request successful |
| 201 Created | Profile created successfully |
| 204 No Content | Deletion successful |
| 400 Bad Request | Invalid data (e.g., duplicate name) |
| 401 Unauthorized | Missing or invalid token |
| 404 Not Found | Profile not found |
| 500 Internal Server Error | Server error |

---

## Design Principles

1. **Simplicity** - Single table design, no complex joins
2. **Flexibility** - JSON storage supports arbitrary fields
3. **Clarity** - Clean, intuitive API endpoints
4. **Reliability** - Proper validation and error handling
5. **Security** - JWT authentication and user isolation

---

### Multi-Environment Setup

Create multiple profiles for different environments, then switch between them:

1. Create profiles for each environment (work, personal, etc.)
2. Switch between them using `PUT /profiles/active`

Example:
```bash
# Create work profile → POST /profiles {"name": "work", ...}
# Create personal profile → POST /profiles {"name": "personal", ...}
# Switch to work → PUT /profiles/active {"profile_name": "work"}
```

### Testing Different Parameters

Create test profiles with different configurations, then easily switch:

1. Create a test profile with specific parameters
2. Switch to test profile
3. Switch back to production when done

Example:
```bash
# Create test profile → POST /profiles {"name": "test-high-temp", "temperature": "0.9", ...}
# Switch to test → PUT /profiles/active {"profile_name": "test-high-temp"}
# Switch back → PUT /profiles/active {"profile_name": "production"}
```

---

### users table (modified)

| Column | Type | Description |
|--------|------|-------------|
| user_id | CHAR(32)/UUID | Primary key |
| username | String | Unique username |
| email | String | Unique email |
| full_name | String | Full name |
| hashed_password | String | Hashed password |
| last_login | DateTime | Last login time |
| is_active | Boolean | Active status |
| is_superadmin | Boolean | Super admin flag |
| **model_configs** | **Text (JSON)** | **Model profiles data** |
| created_at | DateTime | Creation timestamp |
| updated_at | DateTime | Update timestamp |

---

### Recommended: Integration via HTTP API

For security and consistency, other modules should access user profiles through the HTTP API:

```python
import httpx
from typing import Optional, Dict, Any

class UserProfileClient:
    """Client for accessing user profiles via API."""

    def __init__(self, base_url: str = "http://localhost:3080"):
        self.base_url = base_url
        self.client = httpx.AsyncClient()

    async def get_active_profile(
        self,
        user_id: str,
        token: str
    ) -> Optional[Dict[str, Any]]:
        """
        Get the active model profile for a user.

        Args:
            user_id: User UUID
            token: JWT token for authentication

        Returns:
            Profile dict or None if not found
        """
        response = await self.client.get(
            f"{self.base_url}/v3/access/users/{user_id}/profiles/active",
            headers={"Authorization": f"Bearer {token}"}
        )

        if response.status_code == 404:
            return None

        response.raise_for_status()
        return response.json()

    async def close(self):
        await self.client.aclose()


# Usage example
async def get_user_llm_config(user_id: str, token: str) -> Optional[Dict[str, Any]]:
    """Get LLM configuration for a user via API."""
    client = UserProfileClient()

    try:
        profile = await client.get_active_profile(user_id, token)
        if profile:
            return {
                "provider": profile["provider"],
                "model": profile["model"],
                "api_key": profile["api_key"],
                "base_url": profile["base_url"],
                "temperature": profile["temperature"],
            }
        return None
    finally:
        await client.close()
```

**Security Benefits:**
- ✅ Enforces RBAC permission checks
- ✅ Requires valid JWT authentication
- ✅ Maintains audit trail
- ✅ Consistent with external clients

### Direct Repository Access (Not Recommended)

Direct repository access bypasses security controls and should be avoided:

```python
# ⚠️ SECURITY RISK: Skips RBAC, authentication, and audit
from gns3server.db.repositories.users import UsersRepository

async def get_user_config(user_id: str, users_repo: UsersRepository):
    profile = await users_repo.get_active_model_profile(user_id)
    # ... no permission checks!
```

**Use direct repository access only when:**
- The code is already within a secured context with proper validation
- You need to bypass API for performance reasons (rare)
- You're implementing the API layer itself

---

## Best Practices

1. **Use descriptive profile names** - e.g., "work-gpt4", "personal-claude"
2. **Validate API keys** before saving
3. **Keep backups** of important configurations
4. **Test profiles** before using in production
5. **Delete unused profiles** to keep settings organized

---

## Summary

The User Settings API provides a clean, simple approach to configuration management. By storing configurations directly in the users table as JSON, we achieve:

- **Simpler database schema** - No extra tables needed
- **Better performance** - No joins required
- **Easier maintenance** - Less code to manage
- **Same flexibility** - JSON supports arbitrary fields

This design prioritizes simplicity without sacrificing functionality.

**Key Benefits:**
- Single table design eliminates complexity
- Direct field access improves query performance
- Reduced code means easier maintenance
- JSON storage preserves extensibility
- Clean integration with other modules

---

### Quick Reference Card

```
POST   /v3/access/users/{user_id}/profiles              Create profile
GET    /v3/access/users/{user_id}/profiles              List all profiles
GET    /v3/access/users/{user_id}/profiles/active       Get active profile
PUT    /v3/access/users/{user_id}/profiles/active       Switch active profile
PUT    /v3/access/users/{user_id}/profiles/{name}       Update profile
DELETE /v3/access/users/{user_id}/profiles/{name}       Delete profile
```

### Minimum Required Fields

To create a valid profile, you only need:
- `name` - Profile identifier
- `model` - Model name
- `api_key` - API key

All other fields have sensible defaults or are optional.


gns3server/db/models/users.py
```diff
@@ -47,6 +47,7 @@ class User(BaseTable):
     last_login = Column(DateTime)
     is_active = Column(Boolean, default=True)
     is_superadmin = Column(Boolean, default=False)
+    model_configs = Column(Text, nullable=True)  # JSON string for model profiles
```
grossmj, line 50: Shouldn't the column type be 'JSON' instead of 'Text' here?

grossmj:
Looks great, thanks for this! I just have a few questions/observations:

* Would there be a use case for multiple users to have the same model config?
* I could take care to encrypting API keys in the same way as passwords
* I can also add some tests for all your code or do you want to do it?

yueguobin:
> Would there be a use case for multiple users to have the same model config?

**A:** In a multi-user scenario, each user configures their own API key and preferred model. For instance, User A and D might choose the cost-effective DeepSeek, while User B prefers Gemini and User C opts for Grok. Each user operates with their individual credentials. While a single prompt might not perfectly suit every LLM, they are generally compatible enough for this setup to work.

**B:** In another scenario, such as a GNS3-like SaaS platform where the provider supplies the LLM, all users are restricted to a single model chosen by the server. In this case, the configuration can simply be handled via a few key-value pairs in the gns3server config file, making a dedicated user-facing interface unnecessary.

In this PR, I implemented scenario **A**.

> I could take care to encrypting API keys in the same way as passwords

API keys should be stored using encryption; however, they still need to be decrypted when calling the LLM API. Therefore, it is essential to ensure the security of the encryption/decryption keys.

I can try.

> I can also add some tests for all your code or do you want to do it?

I'll try it first, then you can check it out.

gns3server/db/models/users.py
yueguobin: Let me handle it.

yueguobin: Yes, that's it, keep it simple.

yueguobin:
Quick question regarding gns3server's RBAC—is it implemented for collaboration instead of user isolation?

yueguobin:
My bad. If gns3server's RBAC is designed for team collaboration, then this API seems redundant. I assumed it was built for multi-user isolation.

yueguobin:
User Interface  https://github.com/GNS3/gns3-web-ui/pull/1590

grossmj:
> My bad. If gns3server's RBAC is designed for team collaboration, then this API seems redundant. I assumed it was built for multi-user isolation.

The RBAC API is primarily designed for access control (restricting or not what some users can do) and there is also some support for collaboration. That being said, I am not satisfied with the RBAC implementation, the design may not be the best and there are still many untested parts and other bugs/performance issues. For now I recommended to only use users/ groups and be cautious with authorization (roles/ permissions).

https://github.com/GNS3/gns3-server/discussions/1949

grossmj:
Would there be an use case to configure a model / API key for a given group?

I am trying to anticipate what future needs could look like because it is more complicated to change the database schema later on...

yueguobin:
> Would there be an use case to configure a model / API key for a given group?
> 
> I am trying to anticipate what future needs could look like because it is more complicated to change the database schema later on...

There might be system-level, group-level, user-level, and project-level model/API keys — these should cover all the use cases. When implementing the UI, we could allow users to manually select which level of model parameters to use, or default to the priority of Project > User > Group > System.

If we're going to implement all of these, we'll probably need to carefully plan the database schema and API. What do you think?

yueguobin:
I plan to gradually integrate vision models, speech models, or multimodal models in the future (if token costs become more affordable).

  1. Use vision models to recognize topology diagrams, parse them into structured topology descriptions in JSON or YAML format, and rapidly create topologies with basic configurations in projects.
  2. Enable voice capabilities for the copilot through speech models, transforming it into a true lab assistant or learning tutor with conversational voice interaction.
  3. Create attack scenarios where candidates provide video self-introductions followed by AI interrogation, simulating real-world high-pressure work environments.

yueguobin:
Perhaps I could create a dedicated database table to store the LLM configuration information; or use a JSON/YAML configuration file to store this model information, associating the model configurations through user_id/group_id/project_id.

yueguobin:
Add model configs support to user groups.

Perhaps we can use "Thin Backend, Thick Frontend," where the backend provides the interface, and the frontend determines the usage logic.

grossmj:
> There might be system-level, group-level, user-level, and project-level model/API keys — these should cover all the use cases. When implementing the UI, we could allow users to manually select which level of model parameters to use, or default to the priority of Project > User > Group > System.

> If we're going to implement all of these, we'll probably need to carefully plan the database schema and API. What do you think?

Indeed we need to carefully plan the database schema and API. We could only support users and groups for now and if we have a dedicated table for models then it shouldn't be too complicated to also support system. The only potential problem is for projects which don't use the database however I think that users/group & system should cover most use cases.

> Perhaps I could create a dedicated database table to store the LLM configuration information; or use a JSON/YAML configuration file to store this model information, associating the model configurations through user_id/group_id/project_id.

Yes, I definitely think having a dedicated table is the way to go.

> Perhaps we can use "Thin Backend, Thick Frontend," where the backend provides the interface, and the frontend determines the usage logic.

I prefer to have most of the intelligence in the backend actually, one goal of GNS3 v3.0 is to do this, then it is much easier to develop new front ends like a command line client or Python API (https://github.com/GNS3/gns3-server/issues/353) etc.

grossmj:
I am going to review this PR soon, I just got busy on other things ;)

yueguobin:
To be honest, I’m not entirely sure which approach is best yet; I’m still experimenting.

While reading with my child, another idea came to mind. It looks something like this:

**1. Copilot as a standalone service:** Deploy it alongside `gns3server`.

**2. AI Profiles in Web UI:** Use the Copilot API for configuration.

**3. AI Chat in Topology View:** Integrate via the Copilot API as well.

**4. Token Handling:** The Web UI passes the user's JWT token when using AI Profiles or AI Chat. Copilot then uses this token to call `gns3server` APIs to perform various operations.

Let’s keep this PR open for now. I’ll build a demo based on the current changes to provide a more tangible proof-of-concept. Having something visual might give us a clearer sense of how the frontend and backend should ultimately interact.

What do you think?

yueguobin:
Additionally, I plan to store the AI chat data within the project directory.

grossmj:
> Let’s keep this PR open for now. I’ll build a demo based on the current changes to provide a more tangible proof-of-concept. Having something visual might give us a clearer sense of how the frontend and backend should ultimately interact.

Sounds good. I definitely think there is some kind of integration to be done and I also prefer to implement most of the intelligence in the back-end so it will be easier to support for the different frond-ends we have (API, desktop client, web-ui). Thanks for your help 👍

yueguobin:
Close PR, Go to https://github.com/GNS3/gns3-server/pull/2631

PR #2625, opened 2026-03-02
feat(compute): option to disable compute authentication

### LINKED ISSUE #2579: GNS3 3.0.5: Compute authentication cannot be disabled, breaking remote controller connections
**Description:**
```markdown
## Issue Description
After upgrading from GNS3 2.2.54 to 3.0.5, remote controller connections fail with "Host requires authentication (HTTP code 401)". The compute authentication layer cannot be disabled through configuration, preventing GNS3 GUI clients from connecting to the remote server.

## Environment
- **Server:** GNS3 3.0.5 on Ubuntu (GNS3 VM)
- **Client:** GNS3 GUI 3.0.5 on Arch Linux and Windows 11
- **Previous version:** 2.2.54 (working perfectly)
- **Connection:** Remote controller via HTTP on port 80 (nginx proxy to 3080)

## Steps to Reproduce
1. Upgrade GNS3 server from 2.2.54 to 3.0.5
2. Attempt to connect from GNS3 GUI as a remote controller
3. Connection fails with 401 authentication error

## Configuration Attempted
All of the following configurations in `~/.config/GNS3/3.0/gns3_server.conf` were ignored:

**Attempt 1:**
```ini
[Server]
auth = False
```

**Attempt 2:**
```ini
[Server]
auth = False
user = gns3
password = gns3
```

**Attempt 3:**
```ini
[Server]
auth = True
user = gns3
password = <actual_password>
```

**Attempt 4:** Empty config file

## Observed Behavior
Regardless of configuration, server logs always show:
```
WARNING server.py:247 Compute authentication is enabled with username 'gns3' and a randomly generated password. Please set a password in the config file if this compute is to be used by an external controller
```

The `auth = False` setting is completely ignored, and any `password` value set in the config file is also ignored. The compute layer forces authentication with a random password, making it impossible for the controller to connect.

## Expected Behavior
- Setting `auth = False` should disable authentication
- Setting a password in the config should use that password, not generate a random one
- Remote controller connections should work as they did in 2.2.54

## Workaround
Downgrading to GNS3 2.2.55 (latest stable 2.2.x branch) resolves the issue completely.

## Additional Context
This appears related to #2462, but that issue was about Remote Compute connections. This issue affects Remote Controller connections, which is the standard way to connect to a remote GNS3 server.

The fact that GNS3 2.2.55 was released AFTER 3.0.5 suggests this may be a known issue with the 3.0.x branch.
```

markparonyan:
Resolves #2579

### DIFF
--- a/gns3server/api/routes/compute/dependencies/authentication.py
+++ b/gns3server/api/routes/compute/dependencies/authentication.py
@@ -26,12 +26,29 @@ from gns3server.config import Config
 from typing import Optional, Union
 
 log = logging.getLogger(__name__)
-security = HTTPBasic()
+security = HTTPBasic(auto_error=False)
 
 
 def compute_authentication(credentials: Optional[HTTPBasicCredentials] = Depends(security)) -> None:
+    """
+    Authenticate compute requests.
+    
+    Returns None if authentication is disabled or if authentication succeeds
+    Raises HTTPException if authentication is required but credentials are invalid
+    """
 
     server_settings = Config.instance().settings.Server
+
+    if not server_settings.enable_http_auth:
+        return None
+
+    if credentials is None:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Invalid compute username or password",
+            headers={"WWW-Authenticate": "Basic"},
+        )
+
     username = secrets.compare_digest(credentials.username, server_settings.compute_username)
     password = secrets.compare_digest(credentials.password, server_settings.compute_password.get_secret_value())
     if not (username and password):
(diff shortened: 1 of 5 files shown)


markparonyan:
@grossmj Hope I did not break anything

grossmj:
Thanks. You must already be aware but disabling compute authentication means that anyone on the same network could send API requests to the GNS3 server.

markparonyan:
@grossmj Yes, I understand, but sometimes this is a useful option. As I understand this option existed in version 2, why it was stripped out on 3.0 branch in the first place? I guess I should have questioned that before creating a PR =)

PR #2628, opened 2026-03-04
Fix QEMU serial console artifacts by filtering ANSI CPR responses in …

Raizo62:
Hi

On LiveRaizo, with QEMu VM with serial access. The VM has an auto-login on root.
I use GNS3 2.2

When the QEMU VM has fully booted and I only open the console at that point, I see these characters in the prompt: ;1R;80R;80R;80R

I think I've removed all the VM's specifics without solving the problem.

With Copilot's help, we solved the problem by adding this patch.

I preferred to submit a pull request rather than explain the code modification in an issue.

### DIFF
--- a/gns3server/utils/asyncio/telnet_server.py
+++ b/gns3server/utils/asyncio/telnet_server.py
@@ -20,6 +20,7 @@ import socket
 import asyncio
 import asyncio.subprocess
 import struct
+import re
 
 import logging
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


grossmj:
> When the QEMU VM has fully booted and I only open the console at that point, I see these characters in the prompt: ;1R;80R;80R;80R

Thanks for the PR. What terminal program do you use?

Raizo62:
> > When the QEMU VM has fully booted and I only open the console at that point, I see these characters in the prompt: ;1R;80R;80R;80R
> 
> Thanks for the PR. What terminal program do you use?

I use by default `xterm: /usr/bin/xterm`

To seek the bug, i also used the command `netcat 127.0.0.1 5000`

Raizo62:
Sorry, could you remove my fix ?

With this code, we can no longer retrieve the updated console size :-(

Sorry

grossmj:
> Sorry, could you remove my fix ?

No worries, I will revert.

PR #2646, opened 2026-03-26
feat(symbols): add DELETE API for removing custom symbol files

yueguobin:
- Add DELETE /v3/symbols/{symbol_id:path} endpoint
- Prevent deletion of built-in symbols (:/ prefix or in resource directory)
- Clear symbol size cache after deletion
- Require Symbol.Allocate privilege


yueguobin:
https://github.com/user-attachments/assets/d6f8b918-82fb-4c88-ad05-4f64a6f28128

Issue #2691, opened 2026-04-30
500 Internal Server Error when updating compute without password field


yueguobin:
## Bug Description

When updating a compute via the API (PUT /v3/computes/{compute_id}) without providing a password field, the server returns a 500 Internal Server Error.

## Steps to Reproduce

1. Create or have an existing compute registered
2. Send a PUT request to update the compute with a body that does NOT include the password field, for example:
```json
{
  \"host\": \"10.0.0.1\",
  \"port\": 8080
}
```

## Expected Behavior

The compute should be updated successfully with the provided fields, and the password (if already set) should remain unchanged.

## Actual Behavior

The server returns:
```
500 Internal Server Error
```

## Root Cause

In gns3server/db/repositories/computes.py, the update_compute method unconditionally calls compute_update.password.get_secret_value():

```python
update_values[\"password\"] = compute_update.password.get_secret_value()
```

When password is not provided in the request, compute_update.password is None, so calling .get_secret_value() on None raises an AttributeError.

## Suggested Fix

Only set the password in update_values when it is actually provided:

```python
if compute_update.password is not None:
    update_values[\"password\"] = compute_update.password.get_secret_value()
```

## Environment

- Server: GNS3 Server
- Version: 3.10.dev2

PR #2698, opened 2026-05-02
fix: avoid AttributeError when password is not provided in compute update

yueguobin:
## Summary
- Fix AttributeError when updating a compute without providing a password field
- When `compute_update.password` is `None`, calling `.get_secret_value()` on it causes a 500 Internal Server Error
- Only set the password in `update_values` when a password is actually provided


yueguobin:
#2691

Issue #2704, opened 2026-05-05
Excessive reconnection attempts to unreachable compute nodes


yueguobin:
## Description
When a remote compute node is unreachable, the controller makes excessive reconnection attempts, causing log pollution and wasting system resources.

## Symptoms
From the logs, within a 40-second period (00:23:40 to 00:24:20), there are 20+ connection attempts to the same unreachable compute node:

```
00:23:40 INFO Connecting to compute '109764e3-c793-51a3-9b7d-0456cfb81e3a'
00:23:40 WARNING Cannot connect to compute: Cannot connect to host 192.168.100.186:3080
00:23:43 INFO Connecting to compute '109764e3-c793-51a3-9b7d-0456cfb81e3a'
00:23:43 WARNING Cannot connect to compute: Cannot connect to host 192.168.100.186:3080
00:23:43 INFO Connecting to compute '109764e3-c793-51a3-9b7d-0456cfb81e3a'
00:23:43 WARNING Cannot connect to compute: Cannot connect to host 192.168.100.186:3080
... (repeated every 2-3 seconds)
```

Notice the pattern: **every few seconds, there are 2 connection attempts happening almost simultaneously**, indicating multiple reconnection tasks are executing concurrently.

### 1. Multiple Reconnection Triggers
The system has multiple code paths that trigger reconnections:

- **HTTP requests** (`compute.py:344-356`): Every HTTP request to an unconnected compute triggers a connection attempt
- **WebSocket disconnection** (`compute.py:490`): Schedules reconnect after 1 second
- **Connection failure** (`compute.py:393`): Schedules reconnect after 5 seconds

### 2. Reconnection Tasks Accumulate
When multiple concurrent HTTP requests fail:

```python
# Each HTTP request
async def http_query(self, method, path, ...):
    if not self._connected:
        await self.connect()  # Queued by @locking

# Each connect() failure
except ComputeError as e:
    asyncio.get_event_loop().call_later(5, lambda: asyncio.ensure_future(self._try_reconnect()))
    # Creates a new reconnection task!
```

**Flow:**
1. Multiple HTTP requests arrive simultaneously
2. Each calls `await connect()` (queued by `@locking` decorator)
3. Each `connect()` fails and creates a 5-second reconnection task
4. Reconnection tasks accumulate and execute independently
5. **Result: Exponential growth of reconnection tasks**

### 3. No Reconnection Prevention Mechanisms
[...]

PR #2710, opened 2026-05-07
Fix: Check compute connectivity before open() during project deletion

### LINKED ISSUE #2706: Batch project deletion hangs when remote compute is offline
### Description

When batch-deleting multiple projects from the Web UI, if one of the projects uses a remote compute node that is offline, the deletion of the other (local) project may hang indefinitely without returning any HTTP response.

### Steps to Reproduce

1. Configure a remote compute node that is offline/unreachable
2. Create two projects in the Web UI:
   - Project A: uses only the local compute (e.g., IOU devices)
   - Project B: uses a remote compute node that is offline
3. Select both projects and click "Delete"
4. Observe that:
   - Project B returns a 403 Forbidden error with message: `Cannot delete project '...': 1 compute(s) are disconnected: '...'`
   - Project A's DELETE request never returns any response (no 204, no error, no pending)

### Expected Behavior

Both projects should either complete deletion or return an appropriate error response. The failure of one project should not affect the deletion of another independent project.

### Actual Behavior

Project B returns 403 immediately (correct), but Project A's HTTP request hangs indefinitely without any response returned to the client.

### Analysis

The issue was introduced by #2705 which added compute connection status checks before project deletion. When batch-deleting, two DELETE requests are processed concurrently. Although both projects' `project.delete()` methods complete (logs show all steps including close and Web Wireshark cleanup), the HTTP response for the local project is never sent back.

The root cause may be related to concurrent SQLite database access in `rbac_repo.delete_all_ace_starting_with_path()` during the `delete_project` API handler, or a race condition in the async event loop when processing concurrent requests where one fails with a forbidden error.

### Workaround

- Delete projects individually instead of batch-deleting
- Or remove the batch delete button from the Web UI to prevent concurrent deletion requests

### Environment

- GNS3 Server version: 3.0
- OS: Linux
- Web UI: Firefox 149.0
- Database: SQLite (default)

### Related

- #2703 (original issue about long timeouts during project deletion)
- #2705 (fix that added compute connection status checks)

yueguobin:
## Summary

- **Immediate rejection**: When deleting a closed project that references an offline remote compute, the disconnected compute check now runs **before** `open()` by reading the topology file directly (pure file I/O). This avoids a 120-second timeout waiting for the connection attempt to fail.
- **Error logging**: Added proper warning logs in `close()` and `delete_on_computes()` when remote computes are unreachable, instead of silently swallowing errors.
- **Concurrent deletion fix (frontend)**: During investigation of issue #2706, packet capture analysis revealed that when two DELETE requests are sent concurrently and one returns 403, the Angular Web UI cancels the other pending TCP connection, causing the 204 response to never reach the browser. This is a frontend issue — the server correctly deletes the project and sends the response.

### 1. Early compute connectivity check (before `open()`)

**File:** `gns3server/controller/project.py`

Added `_get_disconnected_computes()` method that:
- If project is already opened: uses `self._computes` directly
- If project is closed: reads the topology file (`load_topology`) to extract compute IDs from nodes — pure file I/O, no network connection
- Returns a list of disconnected `Compute` objects

The check is called at the very beginning of `delete()`, before any `open()` attempt. If disconnected computes are found, `ControllerForbiddenError` is raised immediately.

### 2. Error logging improvements

- `close()`: replaced bare `except...pass` with `log.warning` including project ID, compute ID, and error details
- `delete_on_computes()`: wrapped HTTP DELETE in try/except with `log.warning` on failure

## Related Issues

- Closes #2706 — Root cause identified: Angular UI cancels TCP connection on concurrent 403 error (frontend fix needed)
- References #2704 — Deletion path no longer triggers connection attempts to offline computes

## Testing

- 1148 tests passed
- 165 project-specific tests passed
- Verified with Wireshark: disconnected compute rejection is immediate (~10ms vs previous 120s timeout)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,10 @@ venv
 .claude/minmax-settings.json
 .claude/zhipu-settings.json
 .claude/settings.local.json
+.claude/deepseek-settings.json
+.claude/claude-settings.json
+.claude/gemini-settings.json
+.claude/grok-settings.json
 !.claude/development.md  # Exception: allow development docs
 !.claude/skills/         # Exception: allow project skills
 !.claude/memory/         # Exception: allow project memory
@@ -80,3 +84,4 @@ venv
 
 # Tiktoken cache files
 gns3server/agent/gns3_copilot/cache/tiktoken/
+
--- a/gns3server/controller/project.py
+++ b/gns3server/controller/project.py
@@ -1184,6 +1184,16 @@ class Project:
                 if compute_id not in self._computes:
                     self._computes.append(compute_id)
 
+            # Check compute connectivity before creating nodes to avoid
+            # 120-second timeout when a remote compute is unreachable
+            disconnected = self._get_disconnected_computes()
+            if disconnected:
+                compute_names = ", ".join([f"'{c.name}'" for c in disconnected])
+                raise ControllerError(
+                    f"Cannot open project '{self.name}': {len(disconnected)} compute(s) are disconnected: {compute_names}. "
+                    f"Please check the connection and try again."
+                )
+
             for node in topology.get("nodes", []):
                 compute = self.controller.get_compute(node.pop("compute_id"))
                 name = node.pop("name")
--- a/gns3server/controller/project.py
+++ b/gns3server/controller/project.py
@@ -1147,7 +1147,9 @@ class Project:
 
             topology = project_data["topology"]
             for compute in topology.get("computes", []):
-                await self.controller.add_compute(**compute)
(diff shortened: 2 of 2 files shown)


yueguobin:
### Additional fix: Early connectivity check in `open()`

Commit f1b11e7c adds the same early check pattern to `open()`. Before this fix, opening a closed project with nodes on an offline remote compute would block for 120s trying to connect. Now the compute connectivity is checked immediately after loading the topology file and before creating any nodes, providing instant failure with a clear error message.

Related: #2711

PR #2741, opened 2026-05-18
api endpoint implemented

Volo6uev:


yueguobin:
Would you consider bringing this change to the 3.0 branch as well?

Volo6uev:
Sure, I will get on it in a while

PR #2750, opened 2026-05-25
feat: implement simple user isolation based on project ownership

yueguobin:
## Summary

This PR implements a comprehensive RBAC user isolation mechanism for GNS3 server, addressing the gap where users could see all projects regardless of ownership. The implementation uses a three-step permission check that balances ACE-based access control with user ownership and team sharing through resource pools.

### 1. User Isolation (Phase 1)
- **Three-step permission filtering**: ACE check → `created_by` filter → resource pools
- Users can only see projects they created or have been granted access to through resource pools
- **Fixed ACE and resource pool checking logic**: Both project creators (with regular ACEs) and shared users (with resource pool ACEs) can now see shared projects
- Fault-tolerant: Works correctly even with broad ACE configurations (e.g., `path: "/" + propagate: true`)
- Super admins retain full visibility of all projects
- No schema changes required - leverages existing `created_by` field

### 2. Performance Optimization (Batch RBAC Checking)
- **N+1 query elimination**: Replaced per-project `check_user_has_privilege` (N database round-trips) with a single batch method `get_accessible_project_ids` that performs exactly 3 DB queries regardless of project count
- **40x speedup** for large deployments: GET /projects with 10000 projects reduced from ~12s to ~290ms
- **Linear scaling**: Per-project overhead dropped from ~1.2ms to ~0.03ms
- Throughput increased from 823 to 69128 projects/sec
- Maintains identical permission semantics — all ACE path matching, deny evaluation, and resource pool logic preserved in-memory

### 3. Resource Pool Safety
- Prevents deletion of resource pools that are in use by ACE configurations
- Returns detailed error messages showing which users/groups are using the pool
- Resource pools removed from "all endpoints" list - must be explicitly configured
- Prevents accidental access control misconfigurations

### 4. Independent LLMConfig Permissions
- New privileges: `LLMConfig.Audit`, `LLMConfig.Modify`, `LLMConfig.Allocate`
- User-scoped AI profile endpoints now use `LLMConfig.*` permissions
- Added to default User role so regular users can manage their own AI configs
- Group-scoped endpoints retain `Group.*` permissions (admin function)
- Includes Alembic migration for existing databases

## Architecture

The optimized three-step permission check logic:

```python
# Batch ACE + resource pool check (3 DB queries, not N)
all_project_ids = list(controller.projects.keys())
direct_ace_ids, pool_accessible_ids = await rbac_repo.get_accessible_project_ids(
    current_user.user_id, "Project.Audit", all_project_ids
)

# Step 2: Filter direct ACE projects by created_by
result = [p for p in all_projects() 
          if p.id in direct_ace_ids and p.created_by == current_user.username]

# Step 3: Resource pool projects — team sharing (no created_by filter)
result += [p for p in all_projects() if p.id in pool_accessible_ids]
```

**How `get_accessible_project_ids` works**:
1. 3 DB queries upfront: user ACEs, group ACEs, all resources with pool memberships
2. Precompute pool-to-project mapping from the single resources query
3. Check each project's path against user/group ACEs in-memory (pure string matching)
4. Return two sets: directly accessible IDs (needs `created_by` filter) and pool-accessible IDs

**Critical Fix in rbac.py**:
The `check_user_has_privilege` method was updated to use sequential OR logic instead of if-elif logic:
- **Before**: If path is in pool → only check resource pool ACEs, else → only check regular ACEs
- **After**: Check regular ACEs first, then check resource pool ACEs (both can grant access)

This ensures that when a project is shared through a resource pool:
- The project creator (with "All endpoints" ACE) can still see their project
- The shared user (with resource pool ACE) can also see the project

## Design Principles

1. **ACE for basic access control**: Controls whether user can access the system
2. **created_by for user isolation**: Controls which specific resources user can access
3. **Resource pools for project sharing**: The only mechanism for sharing projects between users
4. **No direct ACE sharing**: Users cannot configure ACE directly for specific projects

## Files Changed

- `gns3server/api/routes/controller/projects.py`: Three-step permission filtering + batch RBAC check for GET /projects
- `gns3server/api/routes/controller/pools.py`: Resource pool deletion safety checks
- `gns3server/api/routes/controller/acl.py`: Remove resource pools from endpoint list
- `gns3server/api/routes/controller/llm_model_configs.py`: Use independent LLMConfig permissions
- `gns3server/db/models/privileges.py`: Add LLMConfig privilege definitions
- `gns3server/db/repositories/rbac.py`:
  - Fixed `check_user_has_privilege` to check both regular ACEs and resource pool ACEs (sequential OR logic)
  - Add `get_aces_for_path` method with preloading
  - **Add `get_accessible_project_ids` batch method** for O(1) DB round-trips
- `gns3server/db_migrations/versions/a8829e6c069b_add_llm_config_privileges.py`: Migration
- `tests/stress/benchmark_running_server.py`: Performance benchmark script for GET /projects
- Documentation: Roadmap and design memory files

## Performance Benchmarks

Tested with 10000 projects on a running GNS3 server (user "lisi", full ACE + resource pool configuration):

| Metric | Before (per-project DB queries) | After (batch check, 3 queries) |
|--------|------|-----|
| Average response time | 12155ms | **289ms** |
| Throughput | 823 projects/sec | **69128 projects/sec** |
| Scaling | O(N) DB round-trips | O(1) DB round-trips |

The batch method replaces N×3 DB queries (30000 for 10000 projects) with exactly 3 queries total, regardless of how many projects exist. All path matching and permission evaluation is done in-memory after the initial queries.

## Backward Compatibility

- ✅ Super admins see all projects (unchanged)
- ✅ Resource pool functionality preserved
- ✅ Existing projects and data not affected
- ✅ No breaking changes to API contracts
- ✅ Identical permission semantics — batch method produces same results as per-project checks
- ⚠️ Users with existing broad ACE may see fewer projects (by design)

## Future Work

The roadmap includes 10 phases for future improvements when time permits:
- Phase 2: Template isolation
- Phase 3: Image isolation
- Phase 4: Default ACE for Users group
- Phase 5: ACE architecture refactoring (multi-path support)
- Phase 6: Frontend permission query API
- Phase 7: Resource pool renaming
- Phase 8: Per-user project namespace
- Phase 9: User self-registration
- Phase 10: Email service integration


yueguobin:
I think this interface needs some more tests.

PR #2752, opened 2026-05-25
Api endpoints to manage base configuration files for templates

### LINKED ISSUE #2735: Add API endpoints to manage base configuration files for templates
## Problem Description

Currently, GNS3 server stores references to base configuration files in templates (e.g., VPCS `base_script_file`, IOS `startup_config`), but provides **no API endpoints** to read or edit these configuration file contents.

For example, the VPCS template has a `base_script_file` field that references `vpcs_base_config.txt`. This file is stored in the `configs/` directory and its content is read by the controller when creating nodes (via `controller/node.py:_base_config_file_content()`). However, there is no way for clients to:

- **GET** the current content of these base configuration files
- **PUT/POST** to update the base configuration file content

This means users must manually edit files on the server filesystem to customize base configurations, which is not user-friendly and doesn't integrate with the GNS3 web UI.

## Code Reference

The current implementation in `controller/node.py:231-238`:
```python
def _base_config_file_content(self, path):
    if not os.path.isabs(path):
        path = os.path.join(self.project.controller.configs_path(), path)
    with open(path, encoding="utf-8", errors="ignore") as f:
        return f.read()
```

This method reads files internally but is not exposed via any API endpoint.

## Proposed Solution

Add new API endpoints to the controller template routes:

1. **GET** `/v2/templates/{template_id}/base-config/{filename}` - Retrieve the content of a base configuration file
2. **PUT** `/v2/templates/{template_id}/base-config/{filename}` - Update the content of a base configuration file
3. **GET** `/v2/templates/base-configs` - List all available base configuration files

This would allow the GNS3 web UI to provide an integrated editor for base configuration files, improving the user experience significantly.

## Affected Template Types

- **VPCS**: `base_script_file` (e.g., `vpcs_base_config.txt`)
- **IOS on Unix (IOU)**: `startup_config`, `private_config`
- **Dynamips routers**: `startup_config`, `private_config`
- **Ethernet switches**: potentially `startup_config`

## Use Case

Users want to customize the default VPCS startup script (e.g., to enable DHCP by default, set default IP addresses) without having to SSH into the server and edit text files. An API endpoint would allow the GNS3 web UI to provide a text editor for these files directly in the browser.

## Priority

**Enhancement** - Low priority but would improve UX significantly for users who frequently customize base configurations.

Volo6uev:


Volo6uev:
Hello, I strictly implemented 3 endpoints in api/routes/controller/templates.py, service methods for them and tests. Tests for my code are passing. But tests for pytest tests/api/routes/controller/test_users.py are failing. These failures already existed locally before my changes and are related to authentication/session handling. What should we do?

grossmj:
I am going to review soon. Thanks 👍

PR #2780, opened 2026-06-09
feat: Node file streaming, recursive listing, file type detection, file delete, and MCP file tools

yueguobin:
## Summary

Overhaul of the file handling endpoints for node and project files, adding streaming transfer, recursive listing, file type detection, and file deletion capabilities.

### Streaming
- **File download** (`GET /{node_id}/files/{file_path}`): Stream file content through the controller using `StreamingResponse` instead of buffering entire files in memory
- **File upload** (`POST /{node_id}/files/{file_path}`): Stream request body directly to compute via `request.stream()` without buffering at the controller layer
- Added `stream=True` parameter to `http_query`/`_run_http_query` to return raw aiohttp response for streaming

### File Listing
- **Recursive listing**: Added `recursive` query parameter to node file listing endpoints
- **Subdirectory filtering**: Added `path` query parameter to filter listing to a subdirectory
- **Directories included**: Directory entries are now returned alongside files
- **File type detection**: Replaced simple file extension parsing with `python-magic` for content-based file type identification (`NodeFile.extension` → `NodeFile.file_type`)

### File Deletion
- Added `DELETE /{node_id}/files/{file_path}` endpoint (controller, privilege: Node.Modify)
- Added `DELETE /projects/{project_id}/files/{file_path}` endpoint (compute)
- Both support deleting individual files or entire directories

### MCP Tools (New)
- Added `list_node_files` MCP tool — list files in a node directory with metadata
- Added `get_node_file` MCP tool — read node files with offset/limit line-based pagination (auto-truncate at 50KB)
- Added `write_node_file` MCP tool — write content to node files (create or overwrite)
- Added `delete_node_file` MCP tool — permanently delete node files

### Infrastructure
- Added `params` parameter support to `http_query`/`_run_http_query` for passing query parameters to compute requests
- Added async iterable detection in request data handling for streaming uploads
- Added node file methods to Gns3Connector
- Removed long-commented-out notification polling code

---

### 1. list_node_files

**Function**: List all files/directories in a node directory with metadata (name, size, type, modified time)

**Test scenarios**:
- List files across 5 nodes in a project
- Verify file size, type, and timestamp metadata

**Result**: ✅ Discovered 23 files/directories
- VPCS node: 2 files
- QEMU node: 8 files
- Docker node: 2 directories
- IOU node: 5 files
- Dynamips node: 6 files

---

### 2. get_node_file

**Function**: Read text file content line-by-line with offset/limit pagination

**Test scenarios**:
- Read full config files
- Test various offset/limit parameter combinations
- Verify paginated reading behavior

**Result**: ✅ All pagination scenarios verified

| Parameters | Lines Returned | What Was Tested |
|---|---|---|
| offset=0, limit=1 | Line 1 | Basic read |
| offset=1, limit=1 | Line 2 | Offset read |
| offset=2, limit=2 | Lines 3-4 | Multi-line read |
| offset=4, limit=10 | Remaining content | Past-end handling |
| offset=6, limit=2 | Empty | Out-of-range handling |

**Key features**:
- Large file pagination support
- Safe boundary handling
- Accurate `has_more` status indicator

---

### 3. write_node_file

**Function**: Write content to a node file (create new or overwrite existing)

**Test scenarios**:
- Modify VPCS node startup config
- Set static IP, gateway, DNS parameters

**Result**: ✅ Successfully wrote 130 bytes of new config
- Before: 201 bytes, default config
- After: 130 bytes, custom static IP config
- Name: PC1 → PC-1
- IP: 10.12.1.1/24, Gateway: 10.12.1.254

**Verification**: Started the node and confirmed config via console

---

### 4. delete_node_file

**Function**: Permanently delete a file from a node directory

**Test scenarios**:
- Delete empty vpcs.log file from VPCS node

**Result**: ✅ File deleted and verified
- Before deletion: startup.vpc + vpcs.log (2 files)
- After deletion: startup.vpc + ubridge.log (2 files)
- Verified: vpcs.log no longer exists, node runs normally

---

### Coverage: 100%

- ✅ All 4 tools tested
- ✅ Core functionality verified
- ✅ Edge cases covered (boundary, offset, empty results)
- ✅ Real-world scenarios validated (config management, cleanup)

### DIFF
--- a/gns3server/api/routes/compute/projects.py
+++ b/gns3server/api/routes/compute/projects.py
@@ -194,17 +194,17 @@ async def write_compute_project_file(
     try:
         os.makedirs(os.path.dirname(path), exist_ok=True)
 
-        try:
-            with open(path, "wb+") as f:
-                async for chunk in request.stream():
-                    f.write(chunk)
-        except (UnicodeEncodeError, OSError) as e:
-            pass  # FIXME
+        with open(path, "wb+") as f:
+            async for chunk in request.stream():
+                f.write(chunk)
 
     except FileNotFoundError:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
     except PermissionError:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
+    except OSError as e:
+        log.error(f"Error writing file '{path}': {e}")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
 
 
 @router.delete("/projects/{project_id}/files/{file_path:path}", status_code=status.HTTP_204_NO_CONTENT)
--- a/gns3server/api/routes/compute/projects.py
+++ b/gns3server/api/routes/compute/projects.py
@@ -19,13 +19,14 @@ API routes for projects.
 """
 
 import os
+import shutil
 import urllib.parse
 
 import logging
 
 log = logging.getLogger()
 
-from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
+from fastapi import APIRouter, Depends, HTTPException, Request, Response, status, Query
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import FileResponse
 from typing import List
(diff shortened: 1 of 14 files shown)


yueguobin:
## Docker permissions note

With rootful Docker, volume mount points (e.g. data/, config/) are created as **root:root** on the host side by the Docker daemon when a container starts. This means the GNS3 process cannot write files into mounted directories while the container is running.

**Workaround:** Upload files while the container is stopped, then start the container.

A longer-term fix would involve calling `self._fix_permissions()` in the container start flow to chown host-side directories to the GNS3 user, but this requires further validation.

gns3server/api/routes/compute/projects.py
```diff
@@ -129,59 +130,6 @@ async def delete_compute_project(project: Project = Depends(dep_project)) -> Non
...
-#
-#     :returns: hash
-#     """
-#     stats = {}
-#     # Non blocking call in order to get cpu usage. First call will return 0
-#     stats["cpu_usage_percent"] = CpuPercent.get(interval=None)
-#     stats["memory_usage_percent"] = psutil.virtual_memory().percent
-#     stats["disk_usage_percent"] = psutil.disk_usage(get_default_project_directory()).percent
-#     return {"action": "ping", "event": stats}
-
```
grossmj, line 184: Why did you remove the project notifications?

yueguobin: I saw this code was commented out for a long time, so I removed it.

grossmj: Oh sorry, I hadn't seen the comments :D

PR #2801, opened 2026-07-06
Fix web-wireshark docker build broken by xpra 6.5 release

yueguobin:
## Problem

Since 17 Jun the **Build Docker images** action fails because Xpra 6.5 was released. The Dockerfile pinned xpra with `xpra=6.4*`, which only constrains the `xpra` metapackage. Its sub-packages (`xpra-client`, `xpra-server`, `xpra-codecs`, …) were left unconstrained and default to 6.5, so apt can no longer satisfy the dependency graph:

```
xpra : Depends: xpra-client (= 6.4.4-r0-1) but it is not going to be installed
       Depends: xpra-client-gtk3 (= 6.4.4-r0-1) but it is not going to be installed
       ...
E: Unable to correct problems, you have held broken packages.
```

Tracked in **GNS3/gns3-registry#1044**.

## Fix

Replace the command-line version clamp with APT pinning via a new `/etc/apt/preferences.d/pin-xpra`:

- Pin every `xpra*` package to `6.4*` at priority `1000`.
- Pin `xpra-html5` to `19*` (it follows a separate version scheme; v19 was current alongside xpra 6.4).
- Block every other `xpra*` version with priority `-1`, so if the pinned version ever disappears from the repo apt **fails loudly** instead of silently upgrading to the latest.

The install line drops the `=6.4*` suffix and just installs `xpra` / `xpra-x11`.

This is the APT-pinning approach proposed by @b-ehlers in GNS3/gns3-registry#1044, which is cleaner than enumerating `=6.4*` against every sub-package on the command line.

## Files

- `gns3server/agent/web_wireshark/docker/pin-xpra` — new APT preferences file.
- `gns3server/agent/web_wireshark/docker/Dockerfile` — `COPY` the pin file, drop the `=6.4*` clamp.

Refs: GNS3/gns3-registry#1044


yueguobin:
Closing in favor of a new PR targeting the `3.1` branch, which is the correct base for this fix.

PR #2837, opened 2026-07-28
fix: lowercase GHCR image name in docker-build workflow

my-code-is-grey:
GHCR requires image names to be lowercase. `github.repository_owner` returns the owner's original casing (e.g. `GNS3`), which causes the push to `ghcr.io/GNS3/gns3-server` to fail.

Adds a step to lowercase both the owner and repo name before constructing the GHCR image tag.


my-code-is-grey:
Observed in:

https://github.com/GNS3/gns3-server/actions/runs/29433176905

PR #2840, opened 2026-07-29
Fix unvalidated symlink creation in import_project

grossmj:


candlerb:
I understand the security concern of importing projects from untrusted sources with arbitrary absolute symlinks.

However, I did have a particular use case for this, which won't work any more.  I've raised discussion https://github.com/GNS3/gns3-server/discussions/2841 about how else I might be able to achieve what I'm doing; I'd be grateful for feedback there.

PR #2844, opened 2026-08-04
marker: direction, real-time control, serial-link (WAN) support; uBridge AF_UNIX control channel

yueguobin:
## Summary

gns3-server companion to https://github.com/GNS3/ubridge/pull/116. This advances four coupled areas of the traffic-insight marker feature plus the uBridge control-channel transport, and fixes an IOU serial-capture bug. The uBridge side — the `-U` AF_UNIX socket, `dir=` in match signals, per-filter `enable_packet_filter`, mark-filter survival across `reset_packet_filters`, and the `mark ... linktype` keyword — lands in ubridge#116; this PR is the controller/compute side that drives it.

> Requires the uBridge build from ubridge#116 (≥ 1.2.0), including the `mark ... linktype` keyword for serial-link markers. On an older uBridge the new transport and marker verbs degrade gracefully (see Compatibility).

### 1. uBridge control-channel transport (TCP / AF_UNIX) — unix is now the default

`[Server] ubridge_control_transport = tcp|unix`, **default `unix`**. The server requires uBridge ≥ 1.2.0.

- `unix` (default) → uBridge starts with `-U <path>`: an AF_UNIX control socket authenticated in-kernel via `SO_PEERCRED` (uBridge accepts only its own UID). Sockets are named per node id for self-describing debuggability.
- `tcp` → `-H host:port`, retained for back-compat (uBridge now binds loopback by default).
- If `unix` is selected but the installed uBridge lacks `-U`, the server fails fast with a clear error instead of a silent crash.

### 2. Marker direction

A marker is single-sided: only the chosen capture node's uBridge installs the `mark` filter, yet both directions of the link transit that one bridge. uBridge classifies each match as `tx`/`rx` relative to the capture node and echoes `dir=<tx|rx>` in the `MARK` signal; the server forwards it through to the `marker.match` event so the UI can draw a directed highlight.

- New **create-only** `capture_node_id` pins which endpoint is the observer (else auto-picked in endpoint order); echoed in the marker entry and each signal's `node=<id>`.
- Per-link markers accept `direction: tx|rx|both` (normalized `both`→`null`); an explicit `null` clears it.
- **Definitions reject `tx|rx`** (409): a definition fans out to every link and auto-picks its capture node per link, so a fixed direction has no consistent cross-link meaning. Encode the intended direction in the BPF instead. Direction is restored on project load.

### 3. Real-time marker control (instant `enabled` + per-definition pause/resume)

Previously toggling a marker rebuilt the whole NIO and flushed its pcap. Now:

- **`enabled` drives uBridge directly**: flipping `enabled` sends `enable_packet_filter <bridge> <name> on|off` — no NIO rebuild, no pcap flush; the capture identity and emitted counter are preserved.
- **Per-definition pause/resume**: `POST .../marker-definitions/{name}/pause` and `/resume` toggle every inherited `global-{name}` copy across all links at once. The `paused` flag persists to the `.gns3`, so links created later inherit it already paused.
- Robustness: a toggle is a no-op if the filter isn't installed yet and falls back to an NIO rebuild as a safety net; the bridge map is keyed by `(name, link_id)` so every copy on a multi-link capture node is reachable.

### 4. Serial-link (WAN) marker support

Markers now work on serial links (Cisco HDLC / PPP / Frame Relay / ATM), not just Ethernet. A marker carries a `data_link_type` (default `DLT_EN10MB`); at the uBridge boundary it becomes the `mark ... linktype <dlt>` keyword so the BPF compiles against the right link layer and the pcap is written with the matching linktype.

- The DLT values come straight from `SerialPort.data_link_types` (single source of truth, same as capture): Cisco HDLC → `C_HDLC`, Cisco PPP → `PPP_SERIAL` (DLT 50, the HDLC-framed PPP IOS puts on the wire), Frame Relay → `FRELAY`, ATM → `ATM_RFC1483`. No second mapping table.
- **Per-link markers**: `data_link_type` is create-only; only meaningful for serial links (Ethernet is always EN10MB and the keyword is omitted).
- **Definitions**: `data_link_type` defaults to EN10MB, which makes the definition **Ethernet-only — serial links are skipped** (an EN10MB pcap on a serial link is undecodable). Set a serial DLT to also cover serial links with that encapsulation; Ethernet links stay EN10MB regardless. One definition can serve a mixed topology. Updating `data_link_type` re-fans-out (the set of links covered is re-evaluated).
- **No uBridge rebuild** on a `data_link_type` change — only the affected marker's filter is swapped (`delete_packet_filter` + re-add), reopening its pcap (like a BPF change); sibling markers and the bridge/NIO are untouched.

## API additions

Per-link markers:
- `POST /v3/projects/{pid}/links/{lid}/markers` — accepts `data_link_type` (create-only).
- `PUT /v3/projects/{pid}/links/{lid}/markers/{name}` — instant `enabled` toggle (also updates `bpf/tag/color/highlight_duration`).

Definitions:
- `POST /v3/projects/{pid}/marker-definitions` / `PUT .../{name}` — accept `data_link_type`.
- `POST .../marker-definitions/{name}/pause` and `/resume`.

Compute (marker-capable node types: vpcs, qemu, docker, iou, dynamips, cloud):
- `PUT /adapters/{a}/ports/{p}/markers/{name}` (instant toggle)
- `POST /markers/pause`, `POST /markers/resume`

Full field/endpoint reference: `docs/features/marker-traffic-insight.md`.

## Config

```ini
[Server]
; uBridge control channel transport: "unix" (-U socket_path; AF_UNIX + SO_PEERCRED,
; default, recommended on Linux) or "tcp" (-H host:port).
;ubridge_control_transport = unix
```

## Hardening included

- **IOU serial capture fixed**: the IOU compute `capture/start` route dropped `data_link_type`, so every IOU serial capture was written as Ethernet. Dynamips/cloud/switches already forwarded it; IOU was the only serial-capable node that omitted it. (Also tracked separately as #2845, which is contained in this PR.)
- **IOU marker NIO-update idempotency**: the IOU override of `_ubridge_apply_markers` lacked the incremental guard the generic path has, so adding a private marker to an IOU link that already hosted an inherited `global-*` copy re-added the copy and hit uBridge's duplicate-filter rejection. Fixed to skip already-installed markers, matching the generic path.
- **Per-link `data_link_type` forwarded** on marker create (the route was silently defaulting it to EN10MB).
- **BPF validated once per source**: a definition validates its BPF once at create/update (and once per definition on project load, dropping any whose BPF has gone invalid); the inherited fan-out skips re-validation — creating a definition over *N* links runs one `tcpdump -d` rather than *N*.
- **No empty pcap on restart**: deleting a marker also drops it from the port-NIO cache, so a marker deleted while its node is stopped no longer re-creates an empty pcap when the node later starts.
- **Marker name** `max_length` tightened 128 → 32 (inherited `global-` copies reach ~39).
- **`reset_packet_filters` preserves mark filters**, so unrelated NIO/filter changes no longer close/reopen any marker's pcap.
- **MCP**: `marker_definition` exposes `data_link_type` and no longer exposes `direction` (meaningless for a definition); the per-link `link_marker` tool keeps `direction`.

## Compatibility

All additions are additive. An older uBridge:
- omits `dir=` from match signals → the server falls back to undirected rendering;
- lacks `-U` → set `ubridge_control_transport = tcp`;
- the `enable_packet_filter` / `pause` / `resume` / `mark ... linktype` verbs require the uBridge from ubridge#116 (≥ 1.2.0). Serial-link markers specifically need the `mark ... linktype` keyword.

Target platform: Linux (ubridge#116 drops macOS/Windows).

## Performance & stability

- **Concurrent (bounded) definition fan-out**: a definition's per-link fan-out ran as a serial loop — one compute round-trip per link, so ~1 minute over 500 links. Now `asyncio.gather` + `Semaphore(32)`; links are independent (own `_markers`/`_link_data`), per-link failures stay isolated, and `Project.dump` is synchronous + atomic so concurrent dumps cannot corrupt the topology file. (`apply_defs_to_new_link` stays serial deliberately: all definitions share one link and each push carries the link's full marker set, so concurrent pushes would race.)
- **Batched topology dumps**: every per-link marker op (`start/stop/update_marker`) called `Project.dump()` — a full topology serialize + write. A definition fan-out over 500 links wrote the whole topology 500+ times (each blocking the event loop), which dominated the observed minute. Bulk paths (definition create/update/delete, pause/resume, new-link inheritance) now pass `dump=False` and the caller dumps once after.
- **Open-files limit raised at startup**: every started node holds ~3 file descriptors in the server's table (pidfd + stdout/stderr pipes per child), so a few hundred nodes exhausted the default 1024 soft limit and the uBridge version-check subprocess failed with `EMFILE` ("Too many open files: /dev/null"). At startup, best-effort raise `RLIMIT_NOFILE` to 65535 (capped by the hard limit), before `daemonize()` so the daemon inherits it. Linux-only.

## Tests

~155 new/updated tests across controller, compute, REST, MCP, and the uBridge hypervisor/transport layer, including serial-link fan-out (definition skips serial unless a WAN encapsulation is chosen) and the IOU incremental-apply guard.

### DIFF
--- a/docs/features/marker-traffic-insight.md
+++ b/docs/features/marker-traffic-insight.md
@@ -125,6 +125,28 @@ Because the listener ignores unknown keys, `dir` is **additive**: an older serve
 drops it and an older uBridge simply omits it — either way the system falls back to
 undirected rendering with no error.
 
+### Choosing the capture node
+
+Since `dir` is relative to the capture node, *which* endpoint is the observer decides what
+`tx`/`rx` mean. By default the server auto-picks (first started marker-capable endpoint, in
+link-endpoint order). To pin it — e.g. so `dir=tx` unambiguously means "vpcs1 is sending" —
+pass `capture_node_id` on marker **create**:
+
+```json
+{ "bpf": "icmp", "direction": "tx", "capture_node_id": "<vpcs1 node uuid>" }
+```
+
+The value must be one of the link's two endpoints and a marker-capable type (`vpcs`, `qemu`,
+`docker`, `iou`, `dynamips`, `cloud`); any other id is rejected with `409`. Omit it to keep
+the auto-pick. The chosen id is echoed back as `capture_node_id` in the marker entry and in
+each `MARK` signal's `node=<id>`, so the Web UI always knows the observer regardless of who
+picked it.
+
+`capture_node_id` is **create-only**: it is fixed once the marker exists (changing the
+observer would silently flip the meaning of stored `direction`, so recreate the marker
+instead). It is not accepted on project-level definitions — a definition is link-agnostic and
+has no endpoints to choose from, so inherited markers always auto-pick per link.
+
 ## API Endpoints
 
 All endpoints require a JWT bearer token (`POST /v3/access/users/authenticate`). The
@@ -167,11 +189,16 @@ extra request.
   "name": "icmp",
   "bpf": "icmp",
   "tag": 1,
+  "direction": "tx",
+  "capture_node_id": "a37e2235-e21f-46c9-a2ab-ba0f8c5465e6",
   "color": "#ff5722",
   "highlight_duration": 800,
   "enabled": true
 }
 ```
+
+`direction` and `capture_node_id` are both optional and create-only (see
+[Direction](#direction)).
 
 **Definition create body** (`MarkerDefinitionCreate`, shared by POST and PUT):
 
@@ -224,7 +251,8 @@ extra request.
 | `enabled` | bool | Whether the marker is active |
 | `color` | string \| null | Hex color render hint, e.g. `#ff5722` |
 | `highlight_duration` | int \| null | UI highlight duration in ms after a match; `null` = UI default |
-| `capture_node_id` | string | Server-chosen node whose uBridge hosts the marker |
+| `direction` | string \| null | `tx` / `rx` filter relative to the capture node; `null` = both |
+| `capture_node_id` | string | Node whose uBridge hosts the marker — caller-set on create, else auto-picked |
 | `inherited_from` | string | Source definition name — present on inherited markers only |
 
 ### Definition
--- a/docs/features/marker-traffic-insight.md
+++ b/docs/features/marker-traffic-insight.md
@@ -341,3 +341,12 @@ direction relative to the capture node; see [Direction](#direction).
 - **Persistence.** Definitions and private markers persist in the topology; inherited
   markers are re-created from definitions on project load, so reopening a project restores
   the same configuration and stale inherited copies cannot survive on disk.
+- **Log interpretation across node types.** Each node type logs its startup and link
+  operations differently — do not mistake sparse logs from one type for inactivity.
+  QEMU prints `set_link gns3-<N> on` via its QEMU monitor, which is the most visible
+  startup log among all types. VPCS, Docker, IOU, Dynamips, and Cloud each have their own
+  startup paths (fork + ubridge, container veth, iouyap, Dynamips hypervisor, and TAP
+  device respectively) and none of them emit QEMU-monitor-style logs. To verify marker
+  operations (toggle, pause, resume) on non-QEMU types, either inspect uBridge's
+  own log for `enable_packet_filter` / `marker pause` / `marker resume` commands, or
+  watch the gns3server log for the corresponding compute-route calls at INFO level.
--- a/docs/features/marker-traffic-insight.md
+++ b/docs/features/marker-traffic-insight.md
@@ -147,6 +147,13 @@ observer would silently flip the meaning of stored `direction`, so recreate the 
 instead). It is not accepted on project-level definitions — a definition is link-agnostic and
 has no endpoints to choose from, so inherited markers always auto-pick per link.
 
+For the same reason, a definition **rejects `direction: tx|rx`** (HTTP 409): each inherited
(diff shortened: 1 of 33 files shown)


grossmj:
Should we have `ubridge_control_transport = unix` by default instead?

Also, as discussed on Slack, we should validate that we have the right uBridge version (i.e. >= 1.2.0), it could be when creating a new link or should we do it once when starting the server?

yueguobin:
> Should we have `ubridge_control_transport = unix` by default instead?

Yes, I think having `ubridge_control_transport = unix` by default is good.

> Also, as discussed on Slack, we should validate that we have the right uBridge version (i.e. >= 1.2.0), it could be when creating a new link or should we do it once when starting the server?

This is actually already implemented in `_check_ubridge_version()`  at `gns3server/compute/ubridge/hypervisor.py:155-175` . Updating it to enforce >= 1.2.0 should do it, but I'll make sure to run some concrete tests on it first.

yueguobin:
The current workflow throws an 'uBridge version is too low' error when opening a topology.


https://github.com/user-attachments/assets/a3827357-b1f6-434f-b337-96a01fe41963


```
2026-08-06 21:45:00 ERROR gns3server.api.server:149 Controller error in /v3/projects/20f40515-35b0-4a11-b15f-9552ba6e4715/open (POST): uBridge executable version must be >= 1.2.0
```

PR #2853, opened 2026-08-13
add support basic support for OpenBSD.

hjicks:
there remains some work on ubridge, but this seems to work so far.


grossmj:
I could add this for v2.x however we are moving to a Linux-only back-end for v3.x (and leveraging Linux only features like Linux bridges etc.)

hjicks:
LGTM.

I dont think there is a fundamental difference on bridges (Linux bridge and bridge(4), veb(4) on OpenBSD) far as gns3server is concerned. (ubridge is a bit different. i think there are like two or three functions that need slight modifications). Otherwise stuff looks to much more compatible with linux than what windows or mac do.

PR #2873, opened 2026-09-10
Marker tag aggregate replay driven by resident sharkd sessions

yueguobin:
## Summary

Follow-up to #2869: the tag-keyed aggregate replay — merging the pcaps of every marker sharing a `tag` into one timestamp-ordered timeline once they are all paused — now runs on **resident sharkd sessions** as its only engine. The tshark/PDML path is gone and there is deliberately no degraded mode: without sharkd every replay endpoint returns 501.

## What the Web UI gets

- `GET /v3/projects/{pid}/markers/tags/{tag}/replay/range[?filter=&link=]` — timeline bounds, per-source stats, and the **full merged frame list** (deliberately uncapped; a flat list is the whole contract, rendering cost belongs to the client). Every entry carries Wireshark packet-list columns (`src` / `dst` / `proto` / `info`) plus the coloring-rule hints `bg` / `fg`, so rows render exactly like Wireshark without shipping a colorization engine.
- `GET …/replay/frames?ts=&window_ms=&limit=[&filter=&link=]` — window queries over the merged stream; an empty window is a normal `{"frames": []}`.
- `GET …/replay/frame/detail?ts=&node_id=&link_id=&marker=[&frame_number=]` — lazy single-frame decode: raw hex read straight from the pcap, plus the sharkd protocol tree with keys renamed into a closed, protocol-independent contract (`element` / `label` / `name` / `filter_expr` / `pos` + `size` / `expert` / `generated` / `children`), values untouched. `filter_expr` carries a ready-made display filter with the value baked in (click-to-filter); `pos`/`size` drive click-field → highlight-bytes. The optional `frame_number` disambiguates same-microsecond frames (ts is not unique within a pcap).
- `filter` is a Wireshark display filter applied **before** counting and slicing; invalid expressions are a 400 carrying sharkd's own text (inline-displayable). `link` narrows the stream to one capture source, AND-composing with the filter; `sources` always stays the full tag inventory with engine-free totals so a source dropdown never shrinks.

The unique observable stays the point of the feature: the delta between the same packet hitting two consecutive links is the intermediate node's forwarding latency (host view).

## Engine / session model

One resident `sharkd -` process (line JSON-RPC on stdio) per source pcap:

- lazy spawn; `/tmp` scratch copy with a scratch `HOME` whose preferences **pin the packet-list column layout** — hardened profiles (AppArmor &c.) deny sharkd the project directory, and the column indexes become a contract the server owns rather than an environment default
- per-request `(mtime, size)` validation — a rebuilt/truncated capture never serves stale dissect state
- refcounted sessions with **idle-only LRU eviction** (a tag with more sources than the cap, or a concurrent request, can never have its session killed mid-RPC)
- single-lock check-spawn: concurrent requests share one spawn instead of double-spawning a process nobody reaps
- per-RPC timeout plus **reply-id verification** — a timed-out, dead, malformed or stale reply kills the session for good; a desynchronized session must never serve shifted results
- server shutdown kills every session and drops its scratch copies

The timeline backbone (tag gate, 16-byte-per-frame record-header scan, cross-source merge ordering, hex reads) is plain Python and never depends on the engine.

Full documentation: `docs/features/marker-tag-replay.md`.

## Testing

- 55 replay tests (unit + HTTP routes), real-sharkd coverage where installed: >1000-frame pagination, concurrent single-spawn, in-use eviction protection, timeout/stale-reply aborts, same-microsecond disambiguation, and a closed-set key census that flags any new sharkd tree key
- full suite: 1822 passed
- live validation against a 9-link OSPF project (29 frames, cold range 0.84 s, all four filter regimes, 89-node tree) recorded in the feature doc

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -30,6 +30,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install .[ai-features,dev]
+    - name: Install sharkd (marker replay engine)
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wireshark-common
 
     - name: Lint with flake8
       run: |
--- a/docs/features/marker-tag-replay.md
+++ b/docs/features/marker-tag-replay.md
@@ -124,8 +124,8 @@ without it.
 
 | Method | Path | Description |
 |--------|------|-------------|
-| GET | `/v3/projects/{pid}/markers/tags/{tag}/replay/range[?filter=]` | Timeline metadata + full merged frame list with packet-list columns |
-| GET | `/v3/projects/{pid}/markers/tags/{tag}/replay/frames?ts=&window_ms=&limit=[&filter=]` | Frames with ts in `[T, T+window]`, merged across sources |
+| GET | `/v3/projects/{pid}/markers/tags/{tag}/replay/range[?filter=&link=]` | Timeline metadata + full merged frame list with packet-list columns |
+| GET | `/v3/projects/{pid}/markers/tags/{tag}/replay/frames?ts=&window_ms=&limit=[&filter=&link=]` | Frames with ts in `[T, T+window]`, merged across sources |
 | GET | `/v3/projects/{pid}/markers/tags/{tag}/replay/frame/detail?ts=&node_id=&link_id=&marker=` | Single frame: protocol tree + raw hex (lazy — one call per frame the user opens) |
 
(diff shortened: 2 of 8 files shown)


grossmj:
We need to install wireshark-common package during the Actions workflow.

yueguobin:
> We need to install wireshark-common package during the Actions workflow.

added

Issue #2764, opened 2026-06-02
Docker containers create root-owned files that cannot be cleaned up by GNS3 server


yueguobin:
# Docker containers create root-owned files that cannot be cleaned up by GNS3 server

## Summary

Docker containers managed by GNS3 create files owned by root in the project directory. Since GNS3 server runs as a normal user, it cannot delete these files, causing:

1. **Node deletion fails silently** - directories remain on disk after deleting a node
2. **Project duplication fails** - `shutil.copytree` cannot copy root-owned files, with error:
   ```
   Controller error in /v3/projects/.../duplicate (POST): Cannot duplicate project:
   [('/path/to/source/root', '/path/to/dest/root', "[Errno 13] Permission denied: '/path/to/source/root'")]
   ```
3. **Disk space waste** - orphaned directories accumulate over time

## Environment

- GNS3 server version: 3.1.0.dev3
- OS: Linux (openSUSE)
- Python version: 3.13
- **Note**: This issue has been tested on a local Linux installation. It has **not** been tested in the GNS3 VM environment, where the behavior may be different.

## Steps to Reproduce

1. Create a Docker node using `gns3/webterm:latest` image
2. Start and stop the container
3. Delete the node from the project
4. Check the project directory:
   ```bash
   ls -la /path/to/project/project-files/docker/<uuid>/
   ```
5. Observe that the `root` directory still exists:
   ```
   drwx------ 2 root      root        56  Jun  1 23:55 root
   ```

## Root Cause Analysis

The issue is caused by Docker container configuration in `gns3server/compute/docker/docker_vm.py`:

**Line 476:**
```python
"UsernsMode": "host"  # Container root = Host root
```

**Why this happens:**

1. GNS3 server (running as normal user) creates project directories like `project-files/docker/<uuid>/root`
2. These directories are bind-mounted into the Docker container
3. Container runs as root due to `UsernsMode: "host"` configuration
4. When container processes (running as root) create files in mounted volumes, these files are owned by root on the host (because container UID 0 = host UID 0)
5. GNS3 server (normal user) cannot delete or access these root-owned files

**Example with webterm image:**
```bash
$ docker image inspect gns3/webterm:latest --format='{{.Config.Volumes}}'
map[/root:{}]
```

The `gns3/webterm` image declares `/root` as a volume. When GNS3 creates and mounts this directory, any files created by the container's root process become owned by root on the host.

## The Core Conflict

This is a fundamental architectural issue:

[...]

PR #2874, opened 2026-09-12
fix: reclaim root-owned container files so docker nodes and projects can be deleted

yueguobin:
## Problem

1. https://github.com/GNS3/gns3-server/issues/2764
2. The stop-time permission pass (`_fix_permissions`) necessarily runs **before** the container's processes exit — `docker exec` needs a running container, and the unprivileged server cannot chown root-owned files host-side. Anything the guest writes after that pass — during the shutdown window (syslog archives into `disk1/shutdown`, trace flushes) or after any SIGKILL path — lands as root-owned on the host.

An unprivileged server can neither chown nor delete those files. In practice this makes **node deletion and project deletion fail**:

- `base_node.delete()` → `shutil.rmtree(onerror=set_rw)` — the `set_rw` handler itself EPERMs on root-owned entries
- project deletion → `compute/project.py` and `controller/project.py` both `rmtree` the directory after the nodes close and hit the same wall

This affects every vendor NOS container with persistent volumes (e.g. Cisco XRd persisting `/xr-storage`), where each stop leaves a handful of root-owned files behind.

## Fix

Reclaim the files through the only privilege door a non-root server has: a **one-shot throwaway container** of the node's own image — entrypoint overridden to the GNS3 busybox, so nothing of the guest boots — which chowns the node directory back to the server user (~0.5 s on an XRd control-plane image with an 8k-file volume tree):

```
docker run --rm --network none --pull never \
  --entrypoint /gns3/bin/busybox \
  -v <resources>:/gns3:ro -v <node_dir>:/target \
  <image-id> sh -c 'chmod -R u+rwX … && chown -R <uid>:<gid> …'
```

Two hooks share the same helper:

1. **`close()`** — after the container is removed. This is the load-bearing one: project deletion rmtrees the directory right after the nodes close, so close must leave a clean tree. A quick host-side walk (early-exit, tens of ms) skips the helper entirely when the tree is already clean.
2. **`delete()`** — retry fallback when the rmtree still fails (Docker was down at close time, crash paths), with an error message that names the real cause and gives a manual reclaim command instead of a bare errno.

Robustness details:

- The helper resolves the image by its **create-time image ID** (captured in `_mount_binds`, covers vendor subclasses via `super()`) with `--pull=never`, so a retagged or missing image name cannot turn into a registry pull attempt.
- The chown prefers the image's own coreutils over the static busybox one, mirroring `_fix_permissions` (busybox chown dlopens NSS modules from the image, which mismatch on NOS images).
- `stop()` is untouched — the reclaim only runs on close/delete.

## Testing

- 8 new unit tests in `tests/compute/docker/test_docker_vm.py` (clean/missing-directory skips, foreign-owner detection, full helper command-line assertions, helper failure propagation, close hook, delete retry, delete final error).
- Full docker suite: **210 passed**; shuffled with `--random-order` (two seeds): green.
- Live validation on an XRd control-plane 24.4.1 node: 14 root-owned leftovers reclaimed to 0 in 0.55 s; node and project deletion work afterwards.

### DIFF
--- a/gns3server/compute/docker/docker_vm.py
+++ b/gns3server/compute/docker/docker_vm.py
@@ -1092,10 +1092,14 @@ class DockerVM(BaseNode):
         # as the container-side pass in _fix_permissions). --pull=never keeps
         # a stale tag reference from turning into a registry pull attempt,
         # and the create-time image ID (when known) is immune to retagging.
+        # --user 0:0 overrides a USER baked into the image (e.g.
+        # ghcr.io/nokia/srlinux runs as "user"): without it the "privileged"
+        # helper is exactly as unprivileged as the server itself and
+        # chmod/chown fail with EPERM on files written by other uids.
         image_ref = self._image_id or self._image
         try:
             process = await asyncio.subprocess.create_subprocess_exec(
-                "docker", "run", "--rm", "--network", "none", "--pull", "never",
+                "docker", "run", "--rm", "--network", "none", "--pull", "never", "--user", "0:0",
                 "--entrypoint", "/gns3/bin/busybox",
                 "-v", f"{resources_path}:/gns3:ro",
                 "-v", f"{directory}:/target",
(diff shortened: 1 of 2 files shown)


yueguobin:
Additional validation after the `--user 0:0` fix, on a mixed 10-node topology (2× Cisco IOL-XE, 2× IOL-L2, 2× Nokia SR Linux, 2× webterm, 2× XRd control-plane), all as Docker nodes with persistent volumes:

- All 10 nodes detected foreign-owned leftovers at close (not just vendor NOS — every container runs as root by default, so the shutdown-window tail is universal) and each reclaim completed in ~0.5 s, concurrently, no failures.
- **Project deletion** — the operation this PR fixes — completed cleanly afterwards: node directories gone, zero foreign-owned entries left in the project tree, no one-shot container residue (`--rm`).
- The SR Linux case specifically: `ghcr.io/nokia/srlinux` bakes `USER user:user` into the image and writes files as a large internal uid (229952913), which is what motivated the explicit `--user 0:0`; before it, the helper inherited uid 1000 and chmod/chown failed with EPERM.

PR #2875, opened 2026-09-13
Sync Docker images from the controller to remote computes

yueguobin:
## Problem

Docker node images only exist in each compute\'s own Docker daemon. Images that were loaded or built locally (not available in any registry) cannot be used on a remote compute: `DockerVM.create()` blindly pulls from the Docker repository and node creation fails. Additionally, relying on the image name alone lets a moved tag (e.g. a newer `:latest`) silently serve stale content from a compute that already holds an image under the same name.

## Solution

Docker now reuses the same transparent image sync chain as qemu/iou/dynamips (missing image -> 409 `ImageMissingError` -> controller provides the image -> creation retried):

- **Missing image sync**: when a Docker node is created on a remote compute whose daemon lacks the image, the controller exports it from the Docker daemon on the controller host (`GET /images/{name}/get`) and streams the tar to the compute\'s new `POST /docker/images/load` endpoint, which feeds it to `docker load`. No temp files anywhere. When the image is not available on the controller host either, the compute is asked to pull it from the Docker repository as a fallback (previous behavior, kept for public images).
- **Version consistency**: the controller pins the image id (`Id` from its local daemon) into the create payload. A compute holding a different image under the same tag reports the image as missing, which routes it through the same sync and re-aligns the tag. The pin is omitted when the image is not on the controller host, so registry-only images are unaffected.

The controller host daemon is therefore the source of truth: updating an image there (pull/build) automatically re-aligns every compute on the next node creation.

## Changes

- `compute/docker/docker_vm.py`: raise `ImageMissingError` on a missing (or mismatched) image instead of pulling inline; drop the now unused `DockerVM.pull_image` wrapper; optional `image_digest` constructor field checked in `create()`
- `compute/docker/__init__.py`: `http_query` passes non-dict request bodies through unchanged (chunked) so a tar can be streamed to the daemon; new `load_image()` consuming the `POST /images/load` JSON progress stream
- `api/routes/compute/images.py`: new `POST /docker/images/load` endpoint streaming the tar into the daemon
- `api/routes/compute/docker_nodes.py` + `schemas/compute/docker_nodes.py`: forward the optional `image_digest` create field (compute-internal, validated as `sha256:<hex>`)
- `controller/node.py`: docker branch in `_upload_missing_image` (`_sync_missing_docker_image`: save -> stream -> load, pull fallback, log notifications) and `_add_docker_image_digest` pinning the controller host image id at creation

No new template fields, database columns or migrations.

## Testing

- 259 tests pass in `tests/compute/docker/`, `tests/controller/test_node.py` and `tests/api/routes/compute/test_images.py` (also under `--random-order`): image-missing 409 path, digest match/mismatch, `load_image` success/error streams, load endpoint, controller sync (daemon source + pull fallback) and digest pinning
- Verified on a two-server setup: creating a node from a locally built image (`iol-xe/ioll2-xe`, not in any registry) on a remote compute triggers the missing-image 409, syncs the image in ~13s and the node starts


yueguobin:
https://github.com/user-attachments/assets/c3cc8982-3f69-4aab-9f10-411229fd940b

